PR #66, opened 2020-02-16
Log

kotbegemot:


core/rocketjoe/log/log.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    auto log_t::context(std::shared_ptr<spdlog::async_logger> logger) noexcept -> void {
+        logger_= std::move(logger);
+    }
+
+    auto initialization_logger() -> void{
+        spdlog::set_pattern("[%H:%M:%S %z] [%^%L%$] [thread %t] %v");
+
+        spdlog::init_thread_pool(8192, 1);
+        auto stdout_sink = std::make_shared<spdlog::sinks::stdout_color_sink_mt>();
```
leinlawun, line 32: It seems to me that logs are better to display in stderr, and not in stdout.

Issue #94, opened 2020-05-28
Script mode arguments


leinlawun:
Add the function of forwarding the arguments to the script that is running.

kotbegemot:
https://github.com/pybind/pybind11/issues/2061

PR #100, opened 2020-06-14
Script mode arguments #94

kotbegemot:


apps/rocketjoed/main.cpp
```diff
@@ -193,13 +194,29 @@ int main(int argc, char* argv[]) {
...
+            boost::asio::signal_set sigint_set(env.loop(), SIGINT, SIGTERM);
+            sigint_set.async_wait(
+                [&sigint_set](const boost::system::error_code& /*err*/, int /*num*/) {
+                    sigint_set.cancel();
+                });
+
+            env.startup();
+            break;
+        }
+    }
```
leinlawun: Maybe it’s better to add default behavior so that when adding a new item to the enumeration, the code doesn't break?

kotbegemot: It's a good idea to add a default+ break.

PR #110, opened 2020-07-20
split kernel and zmq network

kotbegemot:


components/python/source/python.cpp
```diff
@@ -0,0 +1,92 @@
...
+        from importlib import import_module
+
+        class Worker(pyrocketjoe.celery.apps.Worker):
+            def __init__(self, app) -> None:
+                super().__init__(app)
+
+        pyrocketjoe.celery.apps.Worker = Worker
+
+        sys.modules['pyrocketjoe'] = pyrocketjoe
+    )__";
```
leinlawun, line 28: Move this into the initialization of the Celery module.

leinlawun (approved):

PR #124, opened 2021-10-15
find v01

makeyavelly:


services/storage/collection.cpp
```diff
@@ -179,8 +186,33 @@ namespace services::storage {
     void collection_t::drop_() {
         storage_.clear();
     }
+
+    std::list<document_t *> collection_t::search_(query_ptr cond) {
```
kotbegemot: why not a vector?

services/storage/collection.cpp
```diff
@@ -179,8 +186,33 @@ namespace services::storage {
...
     }
 
-} // namespace services::storage
\ No newline at end of file
+#ifdef DEV_MODE
+    void collection_t::dummy_insert(document_t &&document) {
+        insert_(document.get_as<std::string>("_id"), std::move(document));
+    }
+
+    std::list<document_t *> collection_t::search_test(query_ptr cond) {
```
kotbegemot: why not a vector?

services/storage/collection.cpp
```diff
@@ -179,8 +186,33 @@ namespace services::storage {
...
+        return search_(std::move(cond));
+    }
+
+    std::list<document_t *> collection_t::find_test(const document_t &cond) {
```
kotbegemot: why not a vector?

services/storage/result.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+#include <list>
+#include "components/storage/document.hpp"
+
+class result_find {
+public:
+    result_find() = delete;
+    result_find(std::list<components::storage::document_t *> &&finded_docs);
```
kotbegemot: why not a vector?

PR #137, opened 2021-11-18
collection drop

makeyavelly:


apps/duck_charmer/wrapper_collection.hpp
```diff
@@ -36,12 +36,15 @@ namespace duck_charmer {
         void remove(py::object cond);
         void delete_one(py::object cond);
         void delete_many(py::object cond);
-        void drop();
+        bool drop();
 
     private:
         const std::string name_;
         wrapper_dispatcher_t* ptr_;
+        wrapper_database* database_;
```
kotbegemot: Why is there a pointer to the base here?

PR #142, opened 2021-12-14
Test v02

makeyavelly:


components/cursor/cursor.cpp
```diff
@@ -20,43 +20,71 @@ namespace components::cursor {
...
-        return data_->size();
+    void cursor_t::sort(std::function<bool(data_t*, data_t*)> sorter) {
+        createListBySort();
+        sorted_.sort(sorter);
+        current_index_ = start_index;
     }
 
-    bool sub_cursor_t::has_next() const {
-        return current_index_ < static_cast<index_t>(size()) - 1;
+    void cursor_t::createListBySort() {
```
kotbegemot: please look at the codestyle in the project

components/document/core/value.hpp
```diff
@@ -96,6 +96,8 @@ namespace document { namespace impl {
         void _retain() const;
         void _release() const;
 
+        template<class TValue> inline TValue as() const;
```
kotbegemot: template not used in place with inline(template is implicitly found inlien )

PR #277, opened 2023-12-12
Add address sanitizer

kotbegemot:
Add address sanitaizers and CI check

### DIFF
--- a/docker/Dockerfile-ubuntu-20
+++ b/docker/Dockerfile-ubuntu-20
@@ -40,7 +40,7 @@ COPY ./CMakeLists.txt ./CMakeLists.txt
 
 WORKDIR /app/build
 
-RUN cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DDEV_MODE=ON  -DSANITIZE_ADDRESS=ON && \
+RUN cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DDEV_MODE=ON && \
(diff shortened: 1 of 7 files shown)


docker/Dockerfile-ubuntu-20
```diff
@@ -40,7 +40,7 @@ COPY ./CMakeLists.txt ./CMakeLists.txt
 
 WORKDIR /app/build
 
-RUN cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DDEV_MODE=ON && \
+RUN cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DDEV_MODE=ON  -DSANITIZE_ADDRESS=ON && \
```
SergeiNA: Is it possible to set it as external parameter? I don't think we should set it as default
`docker build -t [image-name] --build-arg [arg-variable]=[value] .`
https://phoenixnap.com/kb/docker-environment-variables
And just extend readme for test build like:
to build in docker with different sanitizers use this command...

SergeiNA (changes requested):

SergeiNA:
Looks like we have problems with addresses:
https://github.com/duckstax/otterbrix/actions/runs/7178451850/job/19546667024?pr=277
@kotbegemot

SergeiNA (approved):
@kotbegemot Don't forget to create issue to add params into new docker file

PR #270, opened 2023-11-28
results removal

OrangeSmokingJacket:
Functionality of components::result::result_t has been moved to components::cursor::cursor_t.
dispatcher_t and other associated structures now return and receive cursor_t_ptr, which can carry operation status(success or failure), error with description (if there was one)  and changes to a collection

ctests and pytests are working properly

PR #299, opened 2023-12-30
CSharp integration

OrangeSmokingJacket:
1) Name change leftovers
2) Start of a C# integration, which will be expanded later.
Now able to:
Create a database and a collection
Insert, update, delete and select documents via ExecuteSQL()
3) Dockerfiles are now 2 stage builds, which lowers final image size and simplifies dotnet usage


.github/workflows/csharp.yaml
```diff
@@ -1,6 +1,6 @@
 name: dotnet package
 
-#on: [pull_request]
+on: [push]
```
SergeiNA: Why you decide to change it? In other CI builds we use pull_request

.github/workflows/csharp.yaml
```diff
@@ -63,11 +63,10 @@ jobs:
             with:
               dotnet-version: ${{ matrix.dotnet-version }}
 
-          - name: Install dependencies
```
SergeiNA, line 66: @kotbegemot @OrangeSmokingJacket so we don't need those dependences?

docker/Dockerfile-ubuntu-20
```diff
@@ -10,7 +10,14 @@ RUN apt update && \
...
         python3-pip \
         python3-venv \
-        python3-dev curl gnupg apt-transport-https && \
+        python3-dev curl gnupg apt-transport-https \
+        zlib1g \
+        wget && \
+    wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
+    dpkg -i packages-microsoft-prod.deb && \
+    rm packages-microsoft-prod.deb && \
+    apt update && \
```
SergeiNA: Do we really need it here? You used it because of microsoft packages?

docker/Dockerfile-ubuntu-20
```diff
@@ -10,7 +10,14 @@ RUN apt update && \
         ninja-build \
         python3-pip \
         python3-venv \
-        python3-dev curl gnupg apt-transport-https && \
+        python3-dev curl gnupg apt-transport-https \
+        zlib1g \
+        wget && \
+    wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
```
SergeiNA: May be we can wrap all cs dependeces intro one script

docker/Dockerfile-ubuntu-20
```diff
@@ -42,8 +49,26 @@ WORKDIR /app/build
...
+COPY . ./
 
+WORKDIR /app/build
 RUN ctest -C -V --output-on-failure
 
-RUN cd integration/python/ && pytest
\ No newline at end of file
+RUN cd integration/python/ && pytest
+
+RUN cd .. && \
```
SergeiNA: @OrangeSmokingJacket @kotbegemot IMHAO I'd like to see separate docker file for dotnet. Or it should be in each docker file. 
@kotbegemot I think we need to rethink our approach to docker files infrastructure.

components/cursor/cursor.hpp
```diff
@@ -21,16 +21,16 @@ namespace components::cursor {
 
     enum class operation_status_t { success = 1, failure = 0 };
 
-    enum class error_code_t {
+    enum class error_code_t : int32_t {
```
SergeiNA, line 24: Do we really need 32 bits here?

components/cursor/cursor.hpp
```diff
@@ -21,16 +21,16 @@ namespace components::cursor {
...
-        create_phisical_plan_error,
-        other_error
+        database_already_exists = 1,
+        database_not_exists = 2,
+        collection_already_exists = 3,
+        collection_not_exists = 4,
+        collection_dropped = 5,
+        sql_parse_error = 6,
+        create_phisical_plan_error = 7,
+        other_error = -1
```
SergeiNA: If you decided to numerate all enums, standard approach to place default and errors enums at the top. 
But in this case in my opinion use `uint16_t`  and for  `other_error = 1'000`
OR
just place  other_error = -1 at the beginning
```
enum class error_code_t : int16_t {
other = -1,
none = 0,
database_already_exists = 1,
database_not_exists = 2,
....
}
```

components/cursor/cursor.cpp
```diff
@@ -105,7 +105,7 @@ namespace components::cursor {
         : collection_(collection)
         , data_(resource) {}
 
-    void sub_cursor_t::append(data_t data) { data_.push_back(data); }
+    void sub_cursor_t::append(data_t data) { data_.push_back(std::move(data)); }
```
SergeiNA, line 108: Nice!

components/document/document_view.hpp
```diff
@@ -14,6 +14,7 @@ namespace components::document {
 
         document_view_t();
         explicit document_view_t(document_ptr document);
+        explicit document_view_t(document_view_t&& doc_view);
```
SergeiNA, line 17:
```suggestion
        explicit document_view_t(document_view_t&& doc_view) = default;
```
Can we use default move ctor?

SergeiNA: Do we really need r-value ref here?

SergeiNA: Looks like `document_view_t` is a wrapper around `document_ptr` - so might be we can have the same approach as for  `explicit document_view_t(document_ptr document);`?

SergeiNA: Also it might be important to mark it as `noexcept` https://stackoverflow.com/questions/28627348/noexcept-and-copy-move-constructors

services/collection/collection.cpp
```diff
@@ -252,7 +252,7 @@ namespace services::collection {
                              session,
                              make_cursor(context_->resource(), operation_status_t::failure));
         } else {
-            auto result = make_cursor(context_->resource(), operation_status_t::failure);
+            auto result = make_cursor(context_->resource(), operation_status_t::success);
```
SergeiNA, line 255: Why you did this change? What motivation?

integration/cpp/test/test_collection_sql.cpp
```diff
@@ -110,20 +116,23 @@ TEST_CASE("integration::cpp::test_collection::sql::base") {
...
                                                "WHERE count > 90;");
+            REQUIRE(cur->is_success());
             REQUIRE(cur->size() == 9);
         }
         {
             auto session = otterbrix::session_id_t();
             auto cur = dispatcher->execute_sql(session,
                                                "SELECT * FROM TestDatabase.TestCollection "
                                                "WHERE count > 90;");
+            REQUIRE(!cur->is_success());
```
SergeiNA: https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md
```suggestion
            REQUIRE_FALSE(cur->is_success());
```

integration/cpp/test/test_collection_sql.cpp
```diff
@@ -142,20 +152,23 @@ TEST_CASE("integration::cpp::test_collection::sql::base") {
...
                                                "WHERE count < 20;");
+            REQUIRE(cur->is_success());
             REQUIRE(cur->size() == 20);
         }
         {
             auto session = otterbrix::session_id_t();
             auto cur = dispatcher->execute_sql(session,
                                                "SELECT * FROM TestDatabase.TestCollection "
                                                "WHERE count < 20;");
+            REQUIRE(!cur->is_success());
```
SergeiNA:
```suggestion
             REQUIRE_FALSE(cur->is_success());
```

integration/cpp/test/test_collection_ql.cpp
```diff
@@ -112,6 +117,7 @@ TEST_CASE("integration::cpp::test_collection::ql") {
             agg.add_parameter(id_par{1}, 90);
             components::ql::variant_statement_t ql{agg};
             auto cur = dispatcher->execute_ql(session, ql);
+            REQUIRE(!cur->is_success());
```
SergeiNA:
```suggestion
             REQUIRE_FALSE(cur->is_success());
```

integration/cpp/test/test_collection_ql.cpp
```diff
@@ -158,6 +165,7 @@ TEST_CASE("integration::cpp::test_collection::ql") {
             agg.add_parameter(id_par{1}, 20);
             components::ql::variant_statement_t ql{agg};
             auto cur = dispatcher->execute_ql(session, ql);
+            REQUIRE(!cur->is_success());
```
SergeiNA:
```suggestion
             REQUIRE_FALSE(!cur->is_success());
```

integration/c/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 project(c_otterbrix C CXX)
 
 
+set( ${PROJECT_NAME}_headers
```
SergeiNA, line 4: Why you need those changes? Why you decide to change something in C integration?

integration/c/otterbrix.h
```diff
@@ -13,42 +14,126 @@ typedef struct string_view_t {
     size_t size;
 } string_view_t;
 
-///enum level_t { trace, debug, info, warn, err, critical, off, level_nums };
+typedef struct config_t {
```
SergeiNA, line 17: Is this C# integration related?

docker/Dockerfile-ubuntu-20
CLTanuki: https://hub.docker.com/_/microsoft-dotnet-sdk

integration/csharp/src/Duckstax.Otterbrix/Duckstax.Otterbrix.csproj
```diff
@@ -4,6 +4,12 @@
         <TargetFramework>net7.0</TargetFramework>
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
CLTanuki: Do not think, it is necessary, because project successfully builds without it.

integration/csharp/src/Duckstax.Otterbrix/OtterbrixWrapper.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+
+        [DllImport(libotterbrix,
+                   EntryPoint = "otterbrix_create",
+                   ExactSpelling = false,
+                   CallingConvention = CallingConvention.Cdecl)]
+        private static extern IntPtr
+        OtterbrixCreate(TransferConfig config, StringPasser database, StringPasser collection);
+        [DllImport(libotterbrix,
```
CLTanuki, line 108: There should be empty lines between functions. Take on of public editorconfigs like [this](https://github.com/dotnet/roslyn/blob/main/.editorconfig) and add `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` to csproj to avoid having to remember these rules.

integration/csharp/tests/Duckstax.Otterbrix.Tests/WrapperTest.cs
```diff
@@ -0,0 +1,286 @@
...
+            Assert.IsTrue(cursor != null);
+            Assert.IsTrue(cursor.IsSuccess());
+            Assert.IsTrue(cursor.Size() == 100);
+        }
+        {
+            string query = "SELECT * FROM TestDatabase.TestCollection;";
+            CursorWrapper cursor = otterbrix.Execute(query);
+            Assert.IsTrue(cursor != null);
+            Assert.IsTrue(cursor.IsSuccess());
+            Assert.IsFalse(cursor.IsError());
```
CLTanuki, line 34: I think it cannot error if previous assertion passes - it's not a Shroedinger `s cat. Same for other assertions. It can save some CI time.

integration/csharp/tests/Duckstax.Otterbrix.Tests/WrapperTest.cs
```diff
@@ -0,0 +1,286 @@
...
+                Assert.IsTrue(doc != null);
+                Assert.IsTrue(doc.GetLong("count") == 96);
+                Assert.IsTrue(doc.GetString("name") == "Name 96");
+            }
+            {
+                DocumentWrapper doc = cursor.Next();
+                Assert.IsTrue(doc != null);
+                Assert.IsTrue(doc.GetLong("count") == 95);
+                Assert.IsTrue(doc.GetString("name") == "Name 95");
+            }
```
CLTanuki: How about a `for` loop? Possibly You can end with [TestCase parameters](https://www.lambdatest.com/blog/nunit-parameterized-test-examples/) - it's easier to read and maintain.

integration/csharp/tests/Duckstax.Otterbrix.Tests/WrapperTest.cs
```diff
@@ -0,0 +1,286 @@
...
+                Assert.IsTrue(doc != null);
+                Assert.IsTrue(doc.GetLong("count") == 3);
+                Assert.IsTrue(doc.GetString("name") == "Name 3");
+            }
+            {
+                DocumentWrapper doc = cursor.Next();
+                Assert.IsTrue(doc != null);
+                Assert.IsTrue(doc.GetLong("count") == 4);
+                Assert.IsTrue(doc.GetString("name") == "Name 4");
+            }
```
CLTanuki: Same about loop.

.github/workflows/csharp.yaml
OrangeSmokingJacket: I did it for testing, but forgot to change it back

.github/workflows/csharp.yaml
OrangeSmokingJacket: There was only dotnet restore and dotnet test run it implicitly (https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-restore)

docker/Dockerfile-ubuntu-20
OrangeSmokingJacket: It's the only way i managed to make dotnet work properly, but there should be a better way
That's why it is the only docker file i have changed for now

components/cursor/cursor.hpp
OrangeSmokingJacket: It is for C# conversion, but i think it will work with less bytes

components/document/document_view.hpp
OrangeSmokingJacket: So...
Cursor returns const ref to the document_view_t, but if we want to access array or dictionary inside document, it returns new document_view_t as r-value.
And a way to pass it to C# is complicated (You can not pass a class in either direction, but you can access it via pointer). So I store document_view in dynamic memory (otherwise it will go out of scope before C# will get pointer) and pass a pointer, but there is a fork:
1) create move constructor and move created document_view_t to dynamic memory
2) create copy constructor and do the same thing, but it is more complicated

services/collection/collection.cpp
OrangeSmokingJacket: There was a bug there, which i didn't catch in my previous update (#270). When selecting entries, if none were selected, cursor->is_success() should be false, and when something was selected it should become true, but it didn't

integration/c/CMakeLists.txt
OrangeSmokingJacket: C# integration uses C files

integration/c/otterbrix.h
OrangeSmokingJacket: Yes

integration/csharp/tests/Duckstax.Otterbrix.Tests/WrapperTest.cs
OrangeSmokingJacket: The way i build it, it can, there is a case, when IsSuccess() will be false, but IsError() will also be false.
If we select something but nothing was selected, operation won't be successful, because nothing was selected, but also there won't be any errors, because everything worked fine.
It done that way to simplify some checks when there is need to see if something was selected or not

docker/Dockerfile-ubuntu-20
OrangeSmokingJacket: > https://hub.docker.com/_/microsoft-dotnet-sdk

But it is not possible to run docker commands inside a dockerfile. Or is it?

components/cursor/cursor.hpp
OrangeSmokingJacket: For some weird reason, 32 bits are not necessary for C++, but other types behaves strange in C#...
is_error() function has this code: { return error_.type != error_code_t::none; }, and it works fine with all singed and unsigned types, but C#, which just calls that function returns true even if error_.type is "none" for every type, exept int32_t. And i couldn't figure out why...

docker/Dockerfile-ubuntu-20
CLTanuki: Just use it for `FROM` directive at the top and start with the container, full of M$ libs.

.github/workflows/csharp.yaml
SergeiNA: Get it

docker/Dockerfile-ubuntu-20
SergeiNA: In this case it should be a separate dockerfile isn't it?
cc @CLTanuki @OrangeSmokingJacket

docker/Dockerfile-ubuntu-20
SergeiNA: So what thoughts on it?

SergeiNA: So what thoughts on it?

components/cursor/cursor.hpp
SergeiNA: If it's require by c# fine but plz leave a comment like "We have to use  int32_t because of c# comparability type restrictions"

SergeiNA: @OrangeSmokingJacket

services/collection/collection.cpp
SergeiNA: Get it, but in future plz comment such cases in Pr.

docker/Dockerfile-ubuntu-20
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:20.04
+FROM ubuntu:20.04 as builder
```
SergeiNA, line 1: Looks like you use here Multi-stage builds, plz add info about in into PR description. 
Also @kotbegemot PTAL, as more I see cases like this more I want to use separate dockerfile for  C# buuut... We need to discuss it.

docker/Dockerfile-ubuntu-20
```diff
@@ -50,9 +44,17 @@ WORKDIR /app/build
...
     cmake --build . --target all -- -j $(nproc)
 RUN make .
+
+WORKDIR /app/build
+RUN ctest -C -V --output-on-failure
+
+RUN cd integration/python/ && pytest
+
+
+FROM mcr.microsoft.com/dotnet/sdk:7.0
```
SergeiNA, line 52: Leave a few words in comment why we need this

SergeiNA (commented):
From my side we need to discus only dockerfile structure. No more significant notes. So fix all c# issue and merge it.

SergeiNA:
Also I saw 
dotnet package / Run (3.1.x, ubuntu-20.04, 3.8) (pull_request) 
dotnet package / Run (6.0.x, ubuntu-20.04, 3.8) (pull_request)

Is it ok to have both of them? And they both failed.

SergeiNA:
@kotbegemot do we need C# tests for manylinux?

@OrangeSmokingJacket I still see falling
`dotnet package / Run (3.1.x, ubuntu-20.04, 3.8) (pull_request) `
and why you turn off
`dotnet package / Run (6.0.x, ubuntu-20.04, 3.8) (pull_request) `?
Plz add comment.

OrangeSmokingJacket:
> @kotbegemot do we need C# tests for manylinux?
> 
> @OrangeSmokingJacket I still see falling `dotnet package / Run (3.1.x, ubuntu-20.04, 3.8) (pull_request) ` and why you turn off `dotnet package / Run (6.0.x, ubuntu-20.04, 3.8) (pull_request) `? Plz add comment.

There are 2 test cases in each (the first one is from WrapperTest.cs, which works fine, the second is for EntityFramework, which is not implemented yet).
As far as I know, 3.1.x and 6.0.x are running at the same time and failed one cancels the other one

kotbegemot (approved):

PR #301, opened 2024-01-30
[refactoring] remove find_sessoin with name

SergeiNA:
Removing string name field from session_key in dispatcher session. 
Because of further refactoring require remove name usage in create/drop_index  from dispatcher -> require session changes.

### DIFF
--- a/components/session/session.hpp
+++ b/components/session/session.hpp
@@ -14,6 +14,13 @@ namespace components::session {
         bool operator<(const session_id_t& other) const;
         auto data() const -> std::uint64_t;
 
+        /**
+         * @brief Generate new unique session id
+         *
+         * @return session_id_t new unique id
+         */
+        static session_id_t generate_uid();
+
     private:
         std::uint64_t data_;
         std::uint64_t counter_;
--- a/services/disk/manager_disk.cpp
+++ b/services/disk/manager_disk.cpp
@@ -257,10 +257,13 @@ namespace services::disk {
         auto indexes = read_indexes_();
         metafile_indexes_->seek_eof();
         for (const auto& index : indexes) {
+            trace(log_, "manager_disk: load_indexes_ : {}", index.name());
+            // Require to separate sessions for load and create index
+            // For each index create we need to generate unique session id.
             actor_zeta::send(dispatcher,
                              address(),
                              collection::handler_id(collection::route::create_index),
-                             session,
+                             session_id_t::generate_uid(),
                              index,
                              dispatcher);
         }
--- a/services/dispatcher/dispatcher.cpp
+++ b/services/dispatcher/dispatcher.cpp
@@ -27,9 +27,9 @@ namespace services::dispatcher {
                                std::string name)
         : actor_zeta::basic_async_actor(manager_dispatcher, std::move(name))
         , log_(log.clone())
+        , manager_dispatcher_(manager_dispatcher->address())
         , resource_(resource)
         , last_collection_(actor_zeta::base::address_t::empty_address())
-        , manager_dispatcher_(manager_dispatcher->address())
         , memory_storage_(std::move(mstorage))
         , manager_wal_(std::move(mwal))
         , manager_disk_(std::move(mdisk)) {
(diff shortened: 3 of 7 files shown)


services/disk/manager_disk.cpp
```diff
@@ -257,10 +257,11 @@ namespace services::disk {
         auto indexes = read_indexes_();
         metafile_indexes_->seek_eof();
         for (const auto& index : indexes) {
+            trace(log_, "manager_disk: load_indexes_ : {}", index.name());
             actor_zeta::send(dispatcher,
                              address(),
                              collection::handler_id(collection::route::create_index),
-                             session,
```
SergeiNA, line 263: if we pass current session to the create_index later, at the end of load index process this sesson will be removed, but create_index already referenced to it. When we use name for session this wasn't a problem, because find session searches by other key (session + name) which wasn't created yet.

kotbegemot: Here we need to add a comment when we change it so as not to lose it.

services/dispatcher/dispatcher.cpp
```diff
@@ -468,7 +468,7 @@ namespace services::dispatcher {
         collection_full_name_t key(index.database_, index.collection_);
         auto it_collection = collection_address_book_.find(key);
         if (it_collection != collection_address_book_.end()) {
-            make_session(session_to_address_, session_key_t{session, index.name()}, session_t(address, index));
+            make_session(session_to_address_, session_key_t{session, {}}, session_t(address, index));
```
kotbegemot: maybe change make_session(session_to_address_, session_key_t{session, {}}, session_t(address, index)); -> make_session(session_to_address_, session_key_t{session}, session_t(address, index)); ?

services/dispatcher/dispatcher.cpp
```diff
@@ -509,9 +507,7 @@ namespace services::dispatcher {
         collection_full_name_t key(drop_index.database_, drop_index.collection_);
         auto it_collection = collection_address_book_.find(key);
         if (it_collection != collection_address_book_.end()) {
-            make_session(session_to_address_,
-                         session_key_t{session, drop_index.name()},
-                         session_t(address, drop_index));
+            make_session(session_to_address_, session_key_t{session, {}}, session_t(address, drop_index));
```
kotbegemot: maybe change make_session(session_to_address_, session_key_t{session, {}}, session_t(address, drop_index)); -> make_session(session_to_address_, session_key_t{session}, session_t(address, drop_index)); ?

services/disk/manager_disk.cpp
SergeiNA: Done

services/dispatcher/dispatcher.cpp
SergeiNA: I'm trying to do this. Looks like we have collision with multiple definition of make_session from collections. 
I've removed string name from key, but left struct.

components/session/session.hpp
```diff
@@ -14,6 +14,13 @@ namespace components::session {
         bool operator<(const session_id_t& other) const;
         auto data() const -> std::uint64_t;
 
+        /**
+         * @brief Generate new unique session id
+         *
+         * @return session_id_t new unique id
+         */
+        static session_id_t generate_uid();
```
SergeiNA, line 22: explicit naming for new id generation

services/disk/manager_disk.cpp
```diff
@@ -257,10 +257,13 @@ namespace services::disk {
...
         metafile_indexes_->seek_eof();
         for (const auto& index : indexes) {
+            trace(log_, "manager_disk: load_indexes_ : {}", index.name());
+            // Require to separate sessions for load and create index
+            // For each index create we need to generate unique session id.
             actor_zeta::send(dispatcher,
                              address(),
                              collection::handler_id(collection::route::create_index),
-                             session,
+                             session_id_t::generate_uid(),
```
SergeiNA, line 266: Now it looks more self-documented

services/dispatcher/session.hpp
```diff
@@ -36,14 +36,13 @@ class session_t {
     ///components::session::session_id_t session_;
 };
 
+// TODO Remove this session and use logic from collections
+// Move session logic from collections to a separate place
 struct session_key_t {
```
SergeiNA, line 41: Left structure with only one filed - not  a great idea, but otherwise I have to do some general session refactoring.
@kotbegemot

services/disk/manager_disk.cpp
```diff
@@ -257,10 +257,13 @@ namespace services::disk {
         auto indexes = read_indexes_();
         metafile_indexes_->seek_eof();
         for (const auto& index : indexes) {
+            trace(log_, "manager_disk: load_indexes_ : {}", index.name());
+            // Require to separate sessions for load and create index
```
SergeiNA, line 261: @kotbegemot left comment as you proceed.

SergeiNA:
@kotbegemot Can I merge it?

PR #305, opened 2024-02-16
Multicollection physical plan

OrangeSmokingJacket:
Physical plan now supports execution of different collections.

1. planner::create_plan moved from collection to memory_storage
2. physical plan executes in parts (from one split node to the next) in sequence for now. Later a parallel execution will be added.
3. planner takes a "map" of collection names and collection contexts, instead of one context
4. opperators now are intrusive_ptr, instead of unique_ptr


docker/Dockerfile-manylinux2014-wasm
```diff
@@ -98,7 +98,7 @@ RUN   PYBIN=(/opt/python/*/bin) && \
           #${BIN}/python3 -m pip install --no-index --find-links . dist/otterbrix-*.whl  && \
           cp /app/wasm_bin/*.wasm $(echo "./$(${BIN}/python3 get_path.py )/components/wasm_runner/tests/") && \
           cd $(echo "./$(${BIN}/python3 get_path.py )") && \
-          ctest  -C --output-on-failure && \
+          ctest  -C --output-on-failure --timeout 60 && \
```
SergeiNA, line 101: Nice change!

SergeiNA:
@OrangeSmokingJacket Thanks for description but plz separate it into different commits if it's possible next time)
Like
```
commit 1: planner::create_plan moved from collection to memory_storage
commit 2: physical plan executes in parts (from one split node to the next) in sequence for now. Later a parallel execution will be added.
commit 3: planner takes a "map" of collection names and collection contexts, instead of one context
commit 4: opperators now are intrusive_ptr, instead of unique_ptr
```

SergeiNA:
Screenshot 2024-02-18 at 23 08 58
Is this from merge process?

components/logical_plan/node.cpp
```diff
@@ -68,4 +76,11 @@ namespace components::logical_plan {
         , children_(resource)
         , expressions_(resource) {}
 
+    void node_t::collection_dependencies_(std::set<collection_full_name_t>& upper_dependencies) {
+        upper_dependencies.insert(collection_full());
+        for (const auto& child : children_) {
+            child->collection_dependencies_(upper_dependencies);
```
SergeiNA, line 82: @kotbegemot @OrangeSmokingJacket this recursion is ok?

components/logical_plan/node.cpp
```diff
@@ -26,6 +26,14 @@ namespace components::logical_plan {
         std::copy(expressions.begin(), expressions.end(), std::back_inserter(expressions_));
     }
 
+    std::set<collection_full_name_t> node_t::collection_dependencies() {
```
SergeiNA: This function return collection from all nodes and sub nodes?

components/logical_plan/node.cpp
SergeiNA: What if we get some circle dependencies? @OrangeSmokingJacket @kotbegemot

integration/c/main.cpp
```diff
@@ -90,7 +90,6 @@ extern "C" cursor_ptr execute_sql(otterbrix_ptr ptr, string_view_t query_raw) {
     auto session = otterbrix::session_id_t();
     std::string query(query_raw.data, query_raw.size);
     auto cursor = pod_space->space->dispatcher()->execute_sql(session, query);
-    trace(pod_space->space->get_log(), "C++ Error result: {}", cursor->is_error());
```
SergeiNA, line 93: what is this?

services/collection/collection.hpp
```diff
@@ -69,6 +74,7 @@ namespace services::collection {
         */
         components::statistic::statistic_t statistic_;
         storage_t storage_;
+        const collection_full_name_t name_;
```
SergeiNA: not a good idea

services/collection/operators/operator.hpp
```diff
@@ -12,17 +12,19 @@ namespace services::collection {
 
 namespace services::collection::operators {
 
-    enum class operator_type { unused = 0x0, empty, match, insert, remove, update, sort, aggregate };
+    enum class operator_type : uint { unused = 0x0, empty, match, insert, remove, update, sort, aggregate };
```
SergeiNA: How many operator_type can exist? may be we can specify 8 bit type here?
```suggestion
    enum class operator_type : uint8_t { unused = 0x0, empty, match, insert, remove, update, sort, aggregate };
```

services/collection/operators/operator.hpp
```diff
@@ -12,17 +12,19 @@ namespace services::collection {
...
 
     enum class operator_state { created, running, waiting, executed, cleared };
 
-    class operator_t {
+    class operator_t : public boost::intrusive_ref_counter<operator_t> {
```
SergeiNA, line 19: @OrangeSmokingJacket can you remind me why you changed to ref pointers?

services/collection/operators/operator.hpp
```diff
@@ -12,17 +12,19 @@ namespace services::collection {
...
     public:
-        using ptr = std::unique_ptr<operator_t>;
+        using ptr = boost::intrusive_ptr<operator_t>;
 
         operator_t() = delete;
         operator_t(const operator_t&) = delete;
+        operator_t(operator_t&&) = default;
```
SergeiNA, line 25: @OrangeSmokingJacket are you sure this works fine for  `class operator_t : public boost::intrusive_ref_counter<operator_t>` classes?

services/collection/operators/operator.hpp
```diff
@@ -33,6 +35,10 @@ namespace services::collection::operators {
         bool is_executed() const;
         bool is_wait_sync_disk() const;
 
+        const collection_full_name_t& collection_name() const noexcept;
+
+        [[nodiscard]] ptr left() noexcept;
```
SergeiNA: Are you planning to change pointers itself or using underlying class? If you not planning to change pointers (And I don't see it) you can mark those methods as const
```suggestion
        [[nodiscard]] ptr left() const noexcept;
```

OrangeSmokingJacket:
> Screenshot 2024-02-18 at 23 08 58 Is this from merge process?

For some reason github left all previous commits from my csharp intergration pr in history

services/memory_storage/memory_storage.hpp
```diff
@@ -34,9 +37,25 @@ namespace services {
             explicit load_buffer_t(std::pmr::memory_resource* resource);
         };
 
+        struct collection_pack_t {
+            actor_zeta::address_t address;
```
SergeiNA: rename to smth like `collection_address`

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+                      name.to_string(),
+                      session.data());
+                return;
+            }
+            collection_dependencies.emplace(name, collections_.at(name).context);
+        }
         sessions_.emplace(session, session_t{logical_plan, current_message()->sender(), 1});
-        actor_zeta::send(collections_.at(logical_plan->collection_full()),
+        auto plan =
+            collection::planner::create_plan(collection_dependencies, logical_plan, components::ql::limit_t::unlimit());
```
SergeiNA: @OrangeSmokingJacket 
Right now I'm working on index features and I have to place some new information inside physical plan like `mdisk_` and `sessions_` how can do this now? any hints?

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+        context_storage_t collection_dependencies;
+        for (const auto& name : collection_names) {
+            if (!check_collection_(session, name)) {
+                trace(log_,
+                      "memory_storage_t:execute_plan_ collection not found {}, sesion: {}",
+                      name.to_string(),
+                      session.data());
+                return;
+            }
+            collection_dependencies.emplace(name, collections_.at(name).context);
```
SergeiNA: Do we need `name` later? may be we can move it?

services/collection/collection.cpp
```diff
@@ -68,9 +71,9 @@ namespace services::collection {
     }
 
     auto collection_t::execute_plan(const components::session::session_id_t& session,
```
SergeiNA: rename to `execute_sub_plan`

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
                   session.data());
             return;
         }
+        auto collection_names = logical_plan->collection_dependencies();
+        context_storage_t collection_dependencies;
```
SergeiNA: Strange naming:
`auto collection_names = logical_plan->collection_dependencies();` more or less understandable but now I see that `collection_dependencies` might be not the best choice.  
`dependency_tree_collections_names` looks more understandable. 

Also we have
` context_storage_t collection_dependencies;` vs `auto collection_names = logical_plan->collection_dependencies();` - that's mess me up. Let's use 
` context_storage_t collections_context_storage;`

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+                session,
+                make_cursor(default_resource(), error_code_t::create_phisical_plan_error, "invalid query plan"));
+            return;
+        }
+
+        std::stack<collection::operators::operator_ptr> look_up;
+        std::stack<collection::operators::operator_ptr> sub_plans;
+        look_up.push(plan);
+        while (!look_up.empty()) {
+            collection::operators::operator_ptr check_op = look_up.top();
```
SergeiNA:
```suggestion
            auto check_op = look_up.top();
```

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+            collection::planner::create_plan(collection_dependencies, logical_plan, components::ql::limit_t::unlimit());
+
+        if (!plan) {
+            execute_plan_finish_(
+                session,
+                make_cursor(default_resource(), error_code_t::create_phisical_plan_error, "invalid query plan"));
+            return;
+        }
+
+        std::stack<collection::operators::operator_ptr> look_up;
```
SergeiNA, line 309: So here you traverse operation tree and put operation into sub plan from top->right->left order?
Could you move this logic into separate member? 
`auto plans_to_execute = traverse_plan(....);`

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+        std::stack<collection::operators::operator_ptr> sub_plans;
```
SergeiNA, line 310: So all this plans should be executed sequentially?  (`sub_plans`)

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+            }
+            sub_plans.push(look_up.top());
+            look_up.pop();
+            if (check_op != nullptr) {
+                look_up.push(check_op->right());
+                look_up.push(check_op->left());
+            }
+        }
+
+        auto current_plan =
```
SergeiNA, line 329: We set as first operator  leftmost bottom in the tree plan. I can't get how all this stack will be executed, could you plz point me?

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,82 @@ namespace services {
...
+            (plans_.emplace(session, plan_t{std::move(sub_plans), parameters})).first->second.sub_plans.top();
+        actor_zeta::send(collections_.at(current_plan->collection_name()).address,
```
SergeiNA: I see that we send only one operator from the top to the collection.

components/logical_plan/node.cpp
OrangeSmokingJacket: more memory expensive option is to store in each node all previous dependencies, but i`m not sure it is even possible to have working tree with circular dependencies

integration/c/main.cpp
OrangeSmokingJacket: it is when i was trying to figure out what was going on with errors in c# stuff. This trace has 0 value now, and i forgot to remove it previously 
[0cec94b](https://github.com/duckstax/otterbrix/pull/305/commits/0cec94b21c7eadf8c90d9a917620260f8504d97c)

services/collection/operators/operator.hpp
OrangeSmokingJacket: I used it for prototyping, and it is not needed at all now

services/collection/operators/operator.hpp
OrangeSmokingJacket: boost says it is fine
https://www.boost.org/doc/libs/1_84_0/boost/interprocess/smart_ptr/intrusive_ptr.hpp

services/memory_storage/memory_storage.cpp
OrangeSmokingJacket: Add a case for operator_type inside collection::execute_sub_plan, maybe? This section is for collection specifics. It will work for mdisk_.
The same could be done in memory storage in memory_storage_t::execute_plan, for memory_storage stuff

services/memory_storage/memory_storage.cpp
OrangeSmokingJacket: it is returned in a set, and there be structure preserving things happening...
i changed set to unordered set, now should be fine

services/memory_storage/memory_storage.cpp
OrangeSmokingJacket: They were executing sequentially anyway, now with few more calls.
With proper planer later they will be parallel, if possible

services/memory_storage/memory_storage.cpp
OrangeSmokingJacket: It is a start of a sequence. Send sub_plan to execute, then get callback, execute next and so on. I will add a comment for it.

services/memory_storage/memory_storage.cpp
OrangeSmokingJacket: Yeap

OrangeSmokingJacket (commented):
Git again added some weird merge with itself without files for some reason... [401efb2](https://github.com/duckstax/otterbrix/pull/305/commits/401efb2a6cf53e80093912656a5b5fa1e7eb57f9)

services/memory_storage/memory_storage.cpp
```diff
@@ -256,19 +274,92 @@ namespace services {
                   session.data());
             return;
         }
+        auto dependency_tree_collections_names = logical_plan->collection_dependencies();
```
SergeiNA, line 277: I mean rename method `collection_dependencies();` but ok for now.

SergeiNA (approved):

kotbegemot (approved):

PR #327, opened 2024-05-23
Dispatcher - Storage Merge

OrangeSmokingJacket:
Functionality of dispatcher and manager_dispacher is moved to memory_storage (now just storage)


OrangeSmokingJacket:
Currently some sends through actor_zeta are broken:
1) https://github.com/duckstax/otterbrix/blob/8c3c3b45447a119518e2040dc47da99167e7b361/services/disk/manager_disk.cpp#L293
2) https://github.com/duckstax/otterbrix/blob/8c3c3b45447a119518e2040dc47da99167e7b361/services/wal/manager_wal_replicate.hpp#L89

routes, arguments and addresses looks fine, but trace just before send is displayed fine and one after is not called. (whole execution completely freezes on those actor_zeta::send() calls)

PR #348, opened 2024-08-20
B+Tree integration

OrangeSmokingJacket:
Removing RocksDB dependency by integrating native b+tree for document storage and indices on disk


services/disk/metadata.cpp
```diff
@@ -87,9 +88,15 @@ namespace services::disk {
...
 
-    metadata_t::metadata_t(const path_t& file_name)
-        : file_(file_name) {
-        std::string data = file_.readall();
+    metadata_t::metadata_t(local_file_system_t& fs, const path_t& file_name)
+        : file_(open_file(fs,
+                          file_name,
+                          file_flags::WRITE | file_flags::READ | file_flags::FILE_CREATE,
+                          file_lock_type::NO_LOCK)) {
+        char* buffer = new char[file_->file_size()];
```
kotbegemot: Please do it through a std::unique_ptr<char[]>, it will be safer.

kotbegemot (approved):

PR #371, opened 2025-01-26
Readme add info for pip

kotbegemot:


integration/python/otterbrix/pyproject.toml
```diff
@@ -17,3 +17,34 @@ testpaths = ["tests"]
 addopts = [
     "--import-mode=importlib",
 ]
+[project]
+name = "Django"
```
theirix:
```suggestion
name = "otterbrix"
```

theirix (changes requested):

integration/python/otterbrix/pyproject.toml
kotbegemot: fixed it, thanks, didn't notice.

PR #378, opened 2025-02-07
Column storage

OrangeSmokingJacket:


components/table/storage/block_handle.hpp
```diff
@@ -0,0 +1,163 @@
...
+        ART_INDEX = 5,
+        COLUMN_DATA = 6,
+        METADATA = 7,
+        OVERFLOW_STRINGS = 8,
+        IN_MEMORY_TABLE = 9,
+        ALLOCATOR = 10,
+        EXTENSION = 11,
+        TRANSACTION = 12
+    };
+    static constexpr size_t MEMORY_TAG_COUNT = 13;
```
kotbegemot: let's calculate via enum

components/table/concurrent_queue.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
```
kotbegemot: in such implementation it is not a concurrent_queue_t and it is not very suitable for our stage of competitiveness to execution

components/table/data_table.hpp
```diff
@@ -0,0 +1,93 @@
...
+                     std::vector<column_definition_t> column_definitions,
+                     std::string name = "temp");
+        data_table_t(data_table_t& parent, column_definition_t& new_column);
+        data_table_t(data_table_t& parent, uint64_t removed_column);
+        data_table_t(data_table_t& parent,
+                     uint64_t changed_idx,
+                     const types::complex_logical_type& target_type,
+                     const std::vector<storage_index_t>& bound_columns);
+
+        std::vector<types::complex_logical_type> types();
```
kotbegemot: here you might need to use span

kotbegemot (approved):

PR #407, opened 2025-06-06
Dispatcher multithreading bugfixes

OrangeSmokingJacket:
* fixed error when incoming sessions have same hash
* removed redundant second mutex in session_blocker
* fixed error when result was lost or overridden between execute_plan_finish and returning cursor
* fixed dead lock when results returned not in requested order


OrangeSmokingJacket:
There are still problems ones in a while, but mush less often

kotbegemot (approved):

PR #477, opened 2026-04-12
Reading only used columns

tolisso:
Optimization for reading only needed columns from disc. Preserving columns numbers (otherwise all code should be rewritten). Non using columns are being filled with placeholders.


tolisso:
Alternatively, instead of adding projected_cols_ to existing scans, we could create a projected_scan_operator, but it seems like there would be a lot of duplicate code.

OrangeSmokingJacket (approved):

kotbegemot (approved):

PR #486, opened 2026-04-28
catalog migration postgresql style

kotbegemot:
PostgreSQL-style catalog + pipeline-driven resolution

  - pg_* system tables (class/attribute/type/proc/depend/constraint/index/sequence/computed_column), OID-based storage routing, relkind discriminator
  - Catalog reads via pipeline: operator_resolve_* stamps OIDs onto plan nodes; validate/enrich/dispatcher read from the plan tree, async catalog_view removed
  - CHECK / NOT NULL / FK enforcement moved from disk into planner-inserted operators
  - Dynamic schema (relkind='g', Mongo-style): auto-registration of columns on INSERT, versioning, catalog-driven projection
  - WAL: page-structured binary format, per-db_oid routing
  - 12 strong typedefs for SQL identifiers; OID-only DML/DDL routing
  - msgpack → flat-text encoding in the catalog; fast_float for double parsing (libc++ compat)


components/catalog/results/resolve_result.hpp
```diff
@@ -0,0 +1,75 @@
...
+        std::string attname;
+        components::catalog::oid_t atttypid{components::catalog::INVALID_OID};
+        std::int32_t attnum{0};
+        bool attnotnull{false};
+        bool atthasdefault{false};
+        bool attisdropped{false};
+        std::string atttypspec; // serialized complex_logical_type (empty for built-ins)
+        std::string attdefspec; // serialized default value (empty if no default)
+        components::catalog::oid_t attoid{components::catalog::INVALID_OID};
+    };
```
OrangeSmokingJacket, line 37: Could be rearranged a bit to get 112, insted of 120 bytes

components/catalog/results/resolve_result.hpp
```diff
@@ -0,0 +1,75 @@
...
+        components::catalog::oid_t namespace_oid{components::catalog::INVALID_OID};
+        std::string name;
+        std::int32_t pronargs{0};
+        std::uint64_t prouid{0};
+        std::string proargmatchers;
+        std::string prorettype;
+
+        resolve_function_result_t() = default;
+        explicit resolve_function_result_t(std::pmr::memory_resource* /*resource*/) {}
+    };
```
OrangeSmokingJacket, line 77: Could be rearranged a bit to get 120, insted of 128 bytes

components/catalog/builtin_seed.cpp
```diff
@@ -0,0 +1,77 @@
...
+            {wk::public_namespace,              "public"},
+            {wk::information_schema_namespace,  "information_schema"},
+        };
+    }
+
+    std::vector<type_seed_row_t> builtin_type_rows() {
+        return {
+            // Canonical otterbrix names
+            {wk::boolean_type,   "bool"},
+            {wk::int8_type,      "int1"},
```
OrangeSmokingJacket: In our convention it is 'int8', not 'int1' , but I do understand, that that will clash with PostgreSQL int8, that is defined later. I would suggest at least adding a comment for it

components/catalog/cascade_planner.cpp
```diff
@@ -0,0 +1,43 @@
...
+                }
+            }
+            // No external deps → RESTRICT allows the drop (only auto/internal children).
+            return plan;
+        }
+
+        // CASCADE: compute full topological drop order via DFS.
+        // The DFS returns only dependents of the seed; append the seed itself last
+        // so build_drop_sequence deletes its catalog rows after all dependents.
+        try {
```
OrangeSmokingJacket: plan_drop is called in physical_operators, and having try/catch here could be detrimental to performance.
And topological_drop_order could be rewritten, so it does not throw. I will leave a comment there as well

components/catalog/dependency_walker.cpp
```diff
@@ -0,0 +1,44 @@
...
+        };
+
+        void dfs(walk_state& st,
+                 const fetch_deps_fn& fetch_deps,
+                 oid_t cls,
+                 oid_t oid) {
+            if (st.black.count(oid))
+                return;
+            if (st.gray.count(oid))
+                throw cycle_detected_error{oid};
```
OrangeSmokingJacket: It is a void function that throws if there is an error, and that exception is caught later. I would suggest using other error managment technics

components/catalog/catalog_write.hpp
```diff
@@ -0,0 +1,15 @@
...
+#include <components/base/collection_full_name.hpp>
+#include <components/vector/data_chunk.hpp>
+
+namespace components::catalog {
+
+    // Plain data carrier: one pg_catalog row to be appended.
+    // Returned by build_create_table_writes so that catalog/ has no dependency on logical_plan/.
+    struct catalog_write_t {
+        collection_full_name_t table;
+        vector::data_chunk_t   row;
```
OrangeSmokingJacket: vector::data_chunk_t is incredibly inefficient for storing a single row. Could this be done per chunk, instead of per row?

components/catalog/ddl_metadata_builder.cpp
```diff
@@ -0,0 +1,689 @@
...
+            const std::string relkind_str(1, relkind_char);
+            const std::string storagemode_str(1, rk);
+
+            auto chunk = make_pg_row(resource, def->columns,
+                                     [&](vector::data_chunk_t& c, std::pmr::memory_resource* r) {
+                                         c.set_value(0, 0, lv_oid(r, table_oid));
+                                         c.set_value(1, 0, lv_str(r, table_name));
+                                         c.set_value(2, 0, lv_oid(r, namespace_oid));
+                                         c.set_value(3, 0, lv_str(r, relkind_str));
+                                         c.set_value(4, 0, lv_str(r, storagemode_str));
```
OrangeSmokingJacket: logical_value_t is basically a huge variant. Here it is created, passed by const reference and dereferenced. All of that is done to put a value of known type into a column of known type. Please, do not use set_value() and value() in performance critical sections.
Meanwhile, I will work on better setters for data_chunk_t and vector_t.

P.S. Here's a lot of set_value() in this file, but this comment is applicable to all of them.

components/catalog/dependency_walker.hpp
```diff
@@ -0,0 +1,70 @@
...
+    class cycle_detected_error : public std::runtime_error {
+    public:
+        explicit cycle_detected_error(oid_t at)
+            : std::runtime_error("pg_depend cycle detected at oid " + std::to_string(at))
+            , offending_oid_(at) {}
+        oid_t offending_oid() const noexcept { return offending_oid_; }
+
+    private:
+        oid_t offending_oid_;
+    };
```
OrangeSmokingJacket: As I have mentioned before, there is no need to throw 'cycle_detected_error', so class itself is not needed.
Look into core/result_wrapper.hpp

components/catalog/fk_info.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <limits>
+#include <string>
+#include <vector>
+
+namespace components::catalog {
+
+    // FK constraint descriptor resolved at plan-enrich time. Carries enough
+    // information for operator_fk_check_t / operator_fk_cascade_t to enforce the
+    // constraint at execution time without further catalog access.
+    struct fk_info_t {
```
OrangeSmokingJacket, line 16: Some variables could be rearranged to get 256, instead of 264 bytes, which will take 1 cacheline less

components/catalog/helpers.cpp
```diff
@@ -0,0 +1,35 @@
...
+        std::size_t i = 0;
+        while (i < s.size()) {
+            std::size_t j = s.find(',', i);
+            std::string_view tok(s.data() + i, (j == std::string::npos ? s.size() : j) - i);
+            if (!tok.empty()) {
+                try {
+                    out.push_back(static_cast<oid_t>(std::stoul(std::string(tok))));
+                } catch (...) {
+                    // malformed token — skip
+                }
```
OrangeSmokingJacket: std::from_chars should be faster and will not require exceptions catching

components/catalog/system_table_schemas.cpp
```diff
@@ -0,0 +1,751 @@
...
+                    auto entries_str = rest.substr(colon + 1);
+                    std::size_t i = 0;
+                    while (i < entries_str.size()) {
+                        std::size_t comma = entries_str.find(',', i);
+                        std::string token = entries_str.substr(
+                            i, comma == std::string::npos ? std::string::npos : comma - i);
+                        std::size_t eq = token.find('=');
+                        if (eq != std::string::npos) {
+                            std::string label = token.substr(0, eq);
+                            int v = std::stoi(token.substr(eq + 1));
```
OrangeSmokingJacket: Same here, std::stoi() is the only reason to catch exceptions here, but it could be replaced, with std::from_chars

components/catalog/system_table_schemas.cpp
```diff
@@ -0,0 +1,751 @@
...
+                return types::complex_logical_type{LT::UNKNOWN};
+            }
+        }
+        // Flat-text format for all other types.
+        try {
+            size_t pos = 0;
+            return parse_flat_type(resource, spec, pos);
+        } catch (...) {
+            return types::complex_logical_type{LT::UNKNOWN};
+        }
```
OrangeSmokingJacket, line 694: Not a fan of throwing and catching here as well

components/catalog/system_table_schemas.cpp
```diff
@@ -0,0 +1,751 @@
...
+                case LT::DOUBLE:
+                    return types::logical_value_t(resource, std::stod(val_str));
+                case LT::STRING_LITERAL:
+                    return types::logical_value_t(resource, val_str);
+                default:
+                    return std::nullopt;
+            }
+        } catch (...) {
+            return std::nullopt;
+        }
```
OrangeSmokingJacket: Same issue here as with others "from string" calls

components/catalog/system_table_schemas.cpp
```diff
@@ -0,0 +1,755 @@
...
+                                                                //    for ARRAY / STRUCT / UNION / DECIMAL / fixed-width sub-types.
+                                                                //    Empty for builtin scalars (atttypid alone reconstructs the type).
+                                                                //    Mirrors pg_attribute.atttypspec.
+            c.emplace_back("attversion", i64_col(), true);    // 5: was 4
+            c.emplace_back("attrefcount", i64_col(), true);   // 6: was 5
+            return c;
+        }
+    } // namespace
+
+    std::vector<system_table_def_t> all_system_tables() {
```
OrangeSmokingJacket: This function is used a lot, and it creates the same vector all the time. I would suggest making a static variable, or even constexpr, if possible

components/catalog/system_table_schemas.cpp
```diff
@@ -0,0 +1,755 @@
...
+        tables.push_back({"pg_computed_column",  well_known_oid::pg_computed_column_table,   pg_catalog, relkind::regular, pg_computed_column_columns()});
+        tables.push_back({"pg_sequence",         well_known_oid::pg_sequence_table,          pg_catalog, relkind::regular, pg_sequence_columns()});
+        tables.push_back({"pg_rewrite",          well_known_oid::pg_rewrite_table,           pg_catalog, relkind::regular, pg_rewrite_columns()});
+
+        return tables;
+    }
+
+    const system_table_def_t* find_system_table(std::string_view name) {
+        // Static cache: the schemas are immutable once constructed. all_system_tables() rebuilds
+        // the vector each call, so we cache once. Lifetime tied to translation unit.
```
OrangeSmokingJacket: Another verification that all_system_tables() rebuilds the same vector

components/logical_plan/node_drop_index.cpp
```diff
@@ -5,25 +5,40 @@
...
-                                         const std::string& name)
-        : node_t(resource, node_type::drop_index_t, collection)
-        , name_(name) {}
-
-    const std::string& node_drop_index_t::name() const noexcept { return name_; }
+                                         std::string dbname,
+                                         std::string relname,
+                                         std::string indexname,
+                                         std::string schemaname,
+                                         std::string uuid)
```
OrangeSmokingJacket: There are a lot of nodes now that use multiple strings with the same type. I would argue that we need strong types defined for them to prevent future bugs

components/physical_plan/operators/predicates/utils.cpp
```diff
@@ -11,55 +11,41 @@ namespace components::operators::predicates::impl {
...
-                                       size_t) -> core::result_wrapper_t<types::logical_value_t> {
-                auto* vec = chunk_left.at(path);
-                if (!vec->validity().row_is_valid(index_left)) {
-                    return types::logical_value_t(chunk_left.resource(), nullptr);
-                }
-                return vec->value(index_left);
-            };
-        } else {
+        // side_t::right → read from chunk_right (join right side).
+        // side_t::left or undefined → read from chunk_left.
```
OrangeSmokingJacket: Undefined side here means that it was missed at validation step and should be considered an error

components/physical_plan/operators/operator_alter_column_drop.cpp
```diff
@@ -0,0 +1,200 @@
...
+        auto rng = co_await std::move(wf);
+        if (rng.count > 0) ctx->pg_catalog_appends.push_back(std::move(rng));
+
+        // Note: drop_column on a relkind='g' (computing) table is routed to
+        // operator_computed_field_unregister_t in planner.cpp::rewrite_alter_table,
+        // which clears matching pg_computed_column rows. This branch handles
+        // regular (relkind='r') tables only.
+
+        mark_executed();
+    }
```
OrangeSmokingJacket: If something goes wrong, it will be a nightmare to debug that...

components/physical_plan/operators/operator_computed_field_register.cpp
```diff
@@ -0,0 +1,254 @@
...
+                    max_version = v;
+                    latest_atttypid = row[3].is_null()
+                                          ? catalog::INVALID_OID
+                                          : static_cast<catalog::oid_t>(row[3].value<std::uint32_t>());
+                    latest_atttypspec = row[4].is_null()
+                                            ? std::string{}
+                                            : std::string(row[4].value<std::string_view>());
+                    latest_refcount = row[6].is_null() ? 0 : row[6].value<std::int64_t>();
+                }
+            }
```
OrangeSmokingJacket, line 107: Do wee need a row of logical_value_t, if every index is hard-coded?

components/physical_plan/operators/operator_dynamic_cascade_delete.cpp
```diff
@@ -0,0 +1,275 @@
...
+                out.push_back({pg_proc_table,       0});
+                out.push_back({pg_depend_table,     1});
+                out.push_back({pg_depend_table,     3});
+            } else if (classid == pg_namespace_table) {
+                out.push_back({pg_namespace_table,  0});
+                out.push_back({pg_depend_table,     1});
+                out.push_back({pg_depend_table,     3});
+            }
+            return out;
+        }
```
OrangeSmokingJacket: push_back is bad enough on it's own... Can we use 5 static vectors?

components/physical_plan/operators/operator_get_schema.cpp
```diff
@@ -0,0 +1,240 @@
...
+                ordering.reserve(pa_rows.size());
+                for (std::size_t i = 0; i < pa_rows.size(); ++i) {
+                    if (pa_rows[i].size() < 8) continue;
+                    if (!pa_rows[i][7].is_null() && pa_rows[i][7].value<bool>()) {
+                        continue; // attisdropped
+                    }
+                    std::int32_t attnum = pa_rows[i][4].is_null()
+                                              ? 0
+                                              : pa_rows[i][4].value<std::int32_t>();
+                    ordering.emplace_back(attnum, i);
```
OrangeSmokingJacket: This pattern is basically everywhere, so, just to avoid repeating the same arguments everytime:
1) storing data (of known types at compile time) as logical_value_t should be avoided at all cost. (It is super unnecessary both in terms of speed and memory usage).
2) a lot of hard-coded indices and types tied to them.
3) introducing new usage of std::vector, or any other container, especially when std::pmr::memory_resource* is easily available, will make introducing allocation strategies more painful

components/physical_plan/operators/operator_resolve_function.cpp
```diff
@@ -0,0 +1,125 @@
...
+            }
+            // proargmatchers (col 5)
+            if (row.size() > 5 && !row[5].is_null()) {
+                chunk.set_value(5, r, row[5]);
+            }
+            // prorettype (col 6)
+            if (row.size() > 6 && !row[6].is_null()) {
+                chunk.set_value(6, r, row[6]);
+            }
+        }
```
OrangeSmokingJacket, line 149: You know a row index, column index and data type. There is no point of using set_value()

components/physical_plan/operators/operator_resolve_table.cpp
```diff
@@ -0,0 +1,362 @@
...
+                                  ? catalog::INVALID_OID
+                                  : static_cast<catalog::oid_t>(row[1].value<std::uint32_t>());
+                cand.atttypid = row[3].is_null()
+                                    ? catalog::INVALID_OID
+                                    : static_cast<catalog::oid_t>(row[3].value<std::uint32_t>());
+                if (!row[4].is_null()) {
+                    cand.atttypspec.assign(row[4].value<std::string_view>());
+                }
+                cand.attversion  = row[5].value<std::int64_t>();
+                cand.attrefcount = row[6].is_null() ? 0 : row[6].value<std::int64_t>();
```
OrangeSmokingJacket, line 264: Here is a good example for my argument:
You know the result and incomming structs layout, but still there is a slow logical_value_t usage.

components/physical_plan/operators/operator_resolve_table.cpp
```diff
@@ -0,0 +1,362 @@
...
+            out_chunk.set_value(0, i,
+                                 types::logical_value_t(resource_, static_cast<std::int32_t>(i + 1)));
+            out_chunk.set_value(1, i,
+                                 types::logical_value_t(resource_, static_cast<std::uint32_t>(r.attoid)));
+            out_chunk.set_value(2, i,
+                                 types::logical_value_t(resource_, std::string_view{r.attname}));
+            out_chunk.set_value(3, i,
+                                 types::logical_value_t(resource_, static_cast<std::uint32_t>(r.atttypid)));
+            out_chunk.set_value(4, i,
+                                 types::logical_value_t(resource_, std::string_view{r.atttypspec}));
```
OrangeSmokingJacket: Again, you know a layout, both indices, type, type size, but still there is a intermediate logical value created and destroyed for each 'box' in a 5 x 'row_count' matrix

components/table/column_definition.cpp
```diff
@@ -65,4 +66,14 @@ namespace components::table {
 
     void column_definition_t::set_oid(uint64_t oid) { oid_ = oid; }
 
+    void column_definition_t::set_attoid(std::uint32_t v) {
+        if (attoid_ != 0 && attoid_ != v) {
+            std::ostringstream oss;
+            oss << "column_definition_t::set_attoid: attoid is immutable after assignment "
+                << "(column=\"" << name_ << "\", current=" << attoid_ << ", attempted=" << v << ")";
+            throw std::logic_error(oss.str());
```
OrangeSmokingJacket: column_definition_t is used a lot. I would advise against using exceptions here

components/table/constraint.hpp
```diff
@@ -10,13 +10,23 @@ namespace components::table {
...
+        // FK fields — only meaningful when type == FOREIGN_KEY:
+        std::string ref_database;     // referenced table's database (empty = same as child)
+        std::string ref_collection;   // referenced table name
+        std::vector<std::string> ref_columns; // matching column names in referenced table
+        std::string name;             // optional constraint name (autogenerated when empty)
+        // FK semantic flags (PostgreSQL FKCONSTR_* alphabet) — defaults match SQL standard:
+        char fk_matchtype{'s'};       // 's' SIMPLE / 'f' FULL / 'p' PARTIAL
+        char fk_del_action{'a'};      // 'a' NO ACTION / 'r' RESTRICT / 'c' CASCADE / 'n' SET NULL / 'd' SET DEFAULT
+        char fk_upd_action{'a'};      // same alphabet as fk_del_action
     };
```
OrangeSmokingJacket, line 31: Could be rearranged to get 184, instead of 192 bytes, but for now, it is not crutial

components/vector/data_chunk_binary.cpp
```diff
@@ -0,0 +1,432 @@
...
+                }
+            }
+
+            return output;
+        }
+
+        // Read the type header for a single column. Advances scan pointer.
+        types::complex_logical_type read_type_header(const char*& scan, const char* end) {
+            if (scan + 4 > end) {
+                throw std::runtime_error("data_chunk_binary: type header overflows buffer");
```
OrangeSmokingJacket: 1) This exception is not caught anywhere
2) On main branch for a while exists a uniform error handling mechanism (core/result_wrapper.hpp)

services/disk/manager_disk_bootstrap.cpp
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        std::vector<type_seed_row_t> builtin_type_rows() {
+            return {
+                // Canonical otterbrix names
+                {wk::boolean_type,   "bool"},
+                {wk::int8_type,      "int1"},
+                {wk::int16_type,     "int16"},
+                {wk::int32_type,     "int32"},
+                {wk::int64_type,     "int64"},
```
OrangeSmokingJacket: Those are not consistent. I do understand why int1 is not named int8, but at the end of the day... Do we really have to stick with our old names if they are becoming a problem? And PR is already a 30k+ lines...

services/disk/manager_disk_bootstrap.cpp
```diff
@@ -0,0 +1,394 @@
...
+        // already loaded the system ones — here we walk the rest.
+        for (const auto& db_entry : std::filesystem::directory_iterator(config_.path)) {
+            if (!db_entry.is_directory()) continue;
+            const auto db_name = db_entry.path().filename().string();
+            std::uint64_t db_oid_raw = 0;
+            try {
+                db_oid_raw = std::stoull(db_name);
+            } catch (...) {
+                continue; // non-numeric (e.g. wal segment dirs at the same level)
+            }
```
OrangeSmokingJacket: I've already express my feelings about try/catch around std::stoull. Same as before applies here and later as well

OrangeSmokingJacket (changes requested):
This is way too much to grasp in one sitting, but I've tried to sum up my notes on it.
Besides all mentioned in the comments, I would like to see more usage of pmr containers, since we are moving towards a uniform allocation policy. And less logica_value_t usage, because it is slow, and most of the time is not necessary (sure, it is only couple extra CPU cycles [in the best case], but those cycle add up to overall poor performance)

PR #496, opened 2026-05-21
SQL: comma-join + SUBSTRING/LENGTH/REGEXP_REPLACE + LEFT OUTER JOIN regression

kotbegemot:
## Summary

SQL surface extensions to unblock SSB benchmark execution and broader query coverage.

- **G1 comma-join** (`FROM a, b WHERE a.x = b.y`): transformer synthesizes a left-deep tree of binary cross-JoinExpr nodes from N-element fromClause; WHERE filter (sibling `match_t`) recovers inner-join semantics. Validated by SSB q1-1 end-to-end on 5MB lineorder + dim_date.
- **G8/G9 string kernels**: `SUBSTRING` (2/3-arg), `LENGTH`, `REGEXP_REPLACE` row kernels registered via `DEFAULT_FUNCTIONS` uid 5/6/7. Typed accessors `.value<T>()` inline, no `logical_value_t` round-trips. NULL propagation via in-body type check (matchers use `make_always_true` so dispatcher accepts NA inputs).
- **G11 LEFT OUTER JOIN regression**: two `TEST_JOIN` cases covering multi-column ON predicate and ON + WHERE filter combination.
- **`column_pruning` N-ary support**: `process_join` now handles `n>=0` children (binary path unchanged behaviorally for `n==2`; per-child projection split for `n>2`; on-condition walker degrades cleanly for middle children to their own SELECT-list projection). Forward-compatible with future N-ary flattening.

## Deferred

**G13 CREATE VIEW e2e**: parser+transformer ship VIEW writes into `pg_class`/`pg_rewrite`/`pg_depend` (planner.cpp:249-263), but SELECT-time view expansion is not wired across the dispatcher path yet. Parser-level coverage stays in `components/sql/test/test_constraints_ddl.cpp:191-232`.

## Test plan

- [x] Local ctest Debug: 681/681 pass (was 670 baseline + 10 string kernel tests + 1 integration comma_join test)
- [x] Local ctest ASan (Debug + `-fsanitize=address`): 681/681 clean
- [x] SSB q1-1 (`FROM lineorder, dim_date WHERE lo_orderdate=d_datekey AND d_year=1993 AND lo_discount BETWEEN 1 AND 3 AND lo_quantity < 25`) ran end-to-end on 5MB lineorder + dim_date in ~50s — proves G1
- [x] CI: macos, ubuntu, manylinux2014, ASAN runner

### DIFF
--- a/components/compute/kernels/aggregate.cpp
+++ b/components/compute/kernels/aggregate.cpp
@@ -582,6 +582,7 @@ namespace components::compute {
                                             "avg",
                                             "Return data size",
                                             "Results in a single number of the same type as input"));
+        register_string_functions(r);
     }
 
 } // namespace components::compute
--- a/components/compute/kernels/aggregate.cpp
+++ b/components/compute/kernels/aggregate.cpp
@@ -22,6 +22,8 @@ namespace {
         auto operator()(const vector_t& v, size_t count) const {
             auto raw_sum = T();
             for (size_t i = 0; i < count; i++) {
+                if (v.is_null(i))
+                    continue;
                 raw_sum += v.data<T>()[i];
             }
             return logical_value_t{v.resource(), raw_sum};
@@ -30,6 +32,8 @@ namespace {
         auto operator()(const vector_t& v, size_t count) const {
             auto raw_sum = T();
             for (size_t i = 0; i < count; i++) {
+                if (v.is_null(i))
+                    continue;
                 raw_sum += T(v.data<U>()[i]);
             }
             return logical_value_t{v.resource(), raw_sum};
@@ -52,11 +56,33 @@ namespace {
     struct min_operator_t<void> {
         template<typename T>
         auto operator()(const vector_t& v, size_t count) const {
-            return logical_value_t{v.resource(), *std::min_element(v.data<T>(), v.data<T>() + count)};
+            auto* data = v.data<T>();
+            T best = T();
+            bool has_any = false;
+            for (size_t i = 0; i < count; i++) {
+                if (v.is_null(i))
+                    continue;
+                if (!has_any || data[i] < best) {
+                    best = data[i];
+                    has_any = true;
+                }
+            }
+            return logical_value_t{v.resource(), best};
         }
         template<typename T, typename U>
         auto operator()(const vector_t& v, size_t count) const {
-            return logical_value_t{v.resource(), T(*std::min_element(v.data<U>(), v.data<U>() + count))};
+            auto* data = v.data<U>();
+            U best = U();
+            bool has_any = false;
+            for (size_t i = 0; i < count; i++) {
+                if (v.is_null(i))
+                    continue;
+                if (!has_any || data[i] < best) {
+                    best = data[i];
+                    has_any = true;
+                }
+            }
+            return logical_value_t{v.resource(), T(best)};
         }
         template<typename T>
         auto operator()(const logical_value_t& v1, const logical_value_t& v2) const {
@@ -78,11 +104,33 @@ namespace {
     struct max_operator_t<void> {
         template<typename T>
         auto operator()(const vector_t& v, size_t count) const {
-            return logical_value_t{v.resource(), *std::max_element(v.data<T>(), v.data<T>() + count)};
+            auto* data = v.data<T>();
+            T best = T();
+            bool has_any = false;
+            for (size_t i = 0; i < count; i++) {
+                if (v.is_null(i))
+                    continue;
+                if (!has_any || best < data[i]) {
+                    best = data[i];
+                    has_any = true;
+                }
+            }
+            return logical_value_t{v.resource(), best};
         }
         template<typename T, typename U>
         auto operator()(const vector_t& v, size_t count) const {
-            return logical_value_t{v.resource(), T(*std::max_element(v.data<U>(), v.data<U>() + count))};
+            auto* data = v.data<U>();
+            U best = U();
+            bool has_any = false;
+            for (size_t i = 0; i < count; i++) {
+                if (v.is_null(i))
+                    continue;
+                if (!has_any || best < data[i]) {
+                    best = data[i];
+                    has_any = true;
+                }
+            }
+            return logical_value_t{v.resource(), T(best)};
         }
         template<typename T>
         auto operator()(const logical_value_t& v1, const logical_value_t& v2) const {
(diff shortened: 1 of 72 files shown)


components/compute/kernels/aggregate.cpp
OrangeSmokingJacket, line 17: Since you started hanling NULL here (which is a good thing), we might as well do it right.
Per SQL standard, if all values are NULL in sum, min, max and avg, then return value is also NULL
count by definition, counts how many are not NULL, so, it should return a 0

components/compute/kernels/string_functions.cpp
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    // ------------------------------------------------------------------
+    // SUBSTRING(s, start)       — start is 1-based; out-of-range => empty
+    // SUBSTRING(s, start, len)  — both 1-based; len <= 0 => empty; clip to bounds
+    // ------------------------------------------------------------------
+    static core::error_t row_substring_2(kernel_context&,
+                                         const std::pmr::vector<logical_value_t>& inputs,
+                                         std::pmr::vector<logical_value_t>& output) {
+        auto* resource = inputs[0].resource();
```
OrangeSmokingJacket: 'kernel_context' has 'exec_context_t' that has a memory_resource
'inputs' as a pmr container has a resource
'output' as a pmr container has a resource
and all entries in it also have one
I would like to discuss which one should we use.
inputs[0] seems unsafe

components/compute/kernels/string_functions.cpp
```diff
@@ -0,0 +1,218 @@
...
+        // arity::var_args(2) — accept 2 or 3 args; two kernel slots for the overloads.
+        auto fn = std::make_unique<row_function>(name, arity::var_args(2), doc, /*available_kernel_slots=*/2);
+
+        // NA-aware: kernel body propagates NULL, so input matchers stay loose
+        // (otherwise dispatcher rejects NA inputs at signature-match time).
+        kernel_signature_t sig2(function_type_t::row,
+                                {input_type::make_always_true(),
+                                 input_type::make_always_true()},
+                                {output_type::fixed(logical_type::STRING_LITERAL)});
+        row_kernel k2(std::move(sig2), row_substring_2);
```
OrangeSmokingJacket, line 186: In 'row_substring_2' you use hardcoded types for inputs: string, int64, but signature allowes any type. This will be accepted by validate_schema, and than hard crash in funcion itself.
Same in 'row_substring_3' -> If you use fixed types, signature has to except only those types

components/compute/kernels/string_functions.cpp
```diff
@@ -0,0 +1,218 @@
...
+                                                   const std::string& short_doc,
+                                                   const std::string& full_doc) {
+        function_doc doc{short_doc, full_doc, {"string"}, false};
+
+        auto fn = std::make_unique<row_function>(name, arity::unary(), doc, /*available_kernel_slots=*/1);
+
+        kernel_signature_t sig(function_type_t::row,
+                               {input_type::make_always_true()},
+                               {output_type::fixed(logical_type::BIGINT)});
+        row_kernel k(std::move(sig), row_length);
```
OrangeSmokingJacket, line 211: same signature issue again

components/compute/kernels/string_functions.cpp
```diff
@@ -0,0 +1,218 @@
...
+        function_doc doc{short_doc, full_doc, {"string", "pattern", "replacement"}, false};
+
+        auto fn = std::make_unique<row_function>(name, arity::ternary(), doc, /*available_kernel_slots=*/1);
+
+        kernel_signature_t sig(function_type_t::row,
+                               {input_type::make_always_true(),
+                                input_type::make_always_true(),
+                                input_type::make_always_true()},
+                               {output_type::fixed(logical_type::STRING_LITERAL)});
+        row_kernel k(std::move(sig), row_regexp_replace);
```
OrangeSmokingJacket, line 230: same signature issue again

components/physical_plan/operators/scan/full_scan.cpp
```diff
@@ -18,69 +19,116 @@ namespace components::operators {
...
-                        transform_predicate(reinterpret_cast<const expressions::compare_expression_ptr&>(child),
+                    auto child_result =
+                        transform_predicate(resource,
+                                            reinterpret_cast<const expressions::compare_expression_ptr&>(child),
                                             types,
                                             parameters);
-                    if (child_filter) {
-                        filter->child_filters.emplace_back(std::move(child_filter));
+                    if (child_result.has_error()) {
+                        return core::error_t{child_result.error()};
```
OrangeSmokingJacket: It is a good edge case for 'error_t' and 'result_wrapper_t'... If is a design flaw on my end. I will change 'convert_error' method so it will move the error. (It is not modifiable by design, but conversion does not fit into that)

components/physical_plan/operators/scan/full_scan.cpp
```diff
@@ -123,7 +171,14 @@ namespace components::operators {
         auto types = co_await std::move(tf);
 
         // Build filter from expression
-        auto filter = transform_predicate(expression_, types, &ctx->parameters);
+        auto filter_result = transform_predicate(resource_, expression_, types, &ctx->parameters);
+        if (filter_result.has_error()) {
+            set_error(filter_result.error());
+            output_ = make_operator_data(resource_, std::pmr::vector<types::complex_logical_type>{resource_});
+            mark_executed();
+            co_return;
```
OrangeSmokingJacket, line 178: if there is an error, it should have 'operator_state::failed' and no data has to created.
mark_executed sets 'operator_state' to 'executed', possibly hiding the error

components/physical_plan/operators/operator_select.cpp
```diff
@@ -191,7 +191,6 @@ namespace components::operators {
                     break;
                 }
                 case select_column_t::kind::star_expand: {
-                    // Expand all columns from input chunk.
```
OrangeSmokingJacket, line 194: It is not an error, but removing one comment line from a file?

components/sql/transformer/impl/transform_select.cpp
```diff
@@ -176,6 +191,48 @@ namespace components::sql::transform {
...
+                    synth->quals = nullptr; // cross — WHERE supplies the predicate
+                    synth->alias = nullptr;
+                    synth->rtindex = 0;
+                    acc = reinterpret_cast<Node*>(synth);
+                }
+                // Replace the original multi-entry fromClause with a single
+                // top-level JoinExpr so the dispatch below sees T_JoinExpr.
+                node.fromClause->lst.clear();
+                node.fromClause->lst.push_back({acc});
+            }
```
OrangeSmokingJacket, line 239: SQL is a mess... It seems to me like a good way to handle it, but I don't like that new nodes are allocated on our global allocator, instead of arena that is used for every other one. (It is not a direct leak, but they will not be deallocated until main allocator is destroyed)

OrangeSmokingJacket (commented):
Good work!

PR #503, opened 2026-05-26
Hash index added disk index

AnnaBekarevitch:


components/index/disk_hash_single_field_index.hpp
```diff
@@ -0,0 +1,80 @@
...
+        void for_each_pending_insert_impl(uint64_t txn_id,
+                                          const std::function<void(const value_t&, int64_t)>& fn) const final;
+        void for_each_pending_delete_impl(uint64_t txn_id,
+                                          const std::function<void(const value_t&, int64_t)>& fn) const final;
+        void clean_memory_to_new_elements_impl(std::size_t count) final;
+
+        std::unique_ptr<disk_hash_storage_t> disk_table_;
+        mutable result_storage_t scratch_results_;
+        std::unordered_map<uint64_t, std::vector<std::pair<std::string, int64_t>>> pending_inserts_;
+        std::unordered_map<uint64_t, std::vector<std::pair<std::string, int64_t>>> pending_deletes_;
```
OrangeSmokingJacket: We try to use 'pmr' containers as much as possible, and you have access to std::pmr::memory_resource* here

components/index/logical_value_binary_codec.hpp
```diff
@@ -0,0 +1,200 @@
...
+namespace components::index::codec {
+
+    using logical_value_t = components::types::logical_value_t;
+    using logical_type_t = components::types::logical_type;
+
+    template<typename T>
+    inline void append_le(std::pmr::string& out, T v) {
+        unsigned char bytes[sizeof(T)];
+        std::memcpy(bytes, &v, sizeof(T));
+        out.append(reinterpret_cast<const char*>(bytes), sizeof(T));
```
OrangeSmokingJacket, line 20: Not a fan of double copy here, but as a prototype, will do for now

components/index/logical_value_binary_codec.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+    template<typename T>
+    inline void write_le_ptr(uint8_t* p, T v) {
+        std::memcpy(p, &v, sizeof(T));
+    }
+
+    inline void append_logical_value(std::pmr::string& out, const logical_value_t& key) {
+        const auto t = key.type().type();
+        append_le<uint8_t>(out, static_cast<uint8_t>(t));
+        switch (t) {
```
OrangeSmokingJacket: some logical types maps to the same physical type (for example: decimal, timestamp and bigint all could be stored as int64_t).
Storing should remain as logic_type, because you want to preserve the type.
Switch should be done over physical_type, since you are working with representation there.
It will be a bit more supportable in the future.

components/index/disk_hash_single_field_index.cpp
```diff
@@ -0,0 +1,152 @@
...
+            std::pmr::string payload(encoded.data(), encoded.size(), resource);
+            size_t pos = 0;
+            return codec::read_logical_value(resource, payload, pos);
+        }
+    } // namespace
+
+    auto disk_hash_single_field_index_t::insert_impl(value_t key, index_value_t value, core::date::timezone_offset_t local_timezone) -> void {
+        (void) key;
+        (void) value;
+        (void) local_timezone;
```
OrangeSmokingJacket: We use unnamed parameters, in those situation. It is just a style preference

services/index/disk_hash_table.cpp
```diff
@@ -0,0 +1,756 @@
...
+    }
+
+    std::vector<disk_hash_table_t::value_ref_t> disk_hash_table_t::get_all(std::string_view key,
+                                                                            const full_key_loader_t& key_loader) const {
+        std::shared_lock lock(mutex_);
+        const uint32_t key_hash = hash_key(key);
+        uint64_t page_id = bucket_primary_page_id(bucket_id_for_key(key));
+        std::vector<value_ref_t> values;
+
+        std::vector<uint8_t> page(page_size);
```
OrangeSmokingJacket: Ideally, those and other buffers have to be allocated with std::pmr::memory_resource*

CMakeLists.txt
OrangeSmokingJacket, line 1: If we change options for main CMake file, than we also have to change .yaml files and docker files
At least make it ON by default, because we are trying to enforce zero-warning policy where possible

OrangeSmokingJacket (changes requested):
Good change!
I would like to see a bit more allocator aware containers, but everything else is well done

OrangeSmokingJacket (approved):

PR #511, opened 2026-06-09
disk: make manager_disk a pure router/orchestrator; offload per-oid work to agent twins

kotbegemot:
## Summary

`manager_disk_t` becomes a thin **router + fan-out orchestrator + disk-side transaction lifecycle**: it holds no `storages_`, performs no scans/WAL/mutations at runtime, and makes **zero runtime cross-actor synchronous calls**. All per-oid work runs on the `agent_disk_t` twins (keyed by `pool_idx_for_oid`); the remaining `*_sync` helpers are strictly bootstrap-only (pre-scheduler-start, `base_spaces`).

## Changes

**Track A — catalog DDL → agent-0**
- `append_pg_catalog_row` / `delete_pg_catalog_rows` / `update_pg_attribute_commit_id_field` / `compact_relkind_g_storage` / `mark_storage_dropped` moved to agent `_inner` handlers (scan + WAL + mutation on the agent thread); manager handlers are thin routers.
- WAL wiring: `agent_disk_t::set_manager_wal_sync` + `manager_wal_addr_`, fanned out in `manager_disk_t::sync`; catalog WAL is now written on agent-0.
- Removed all post-bootstrap `storage_entry_sync` borrows (fixes the shared-state rule).

**Track B — catalog reads**
- Offloaded `read_chunks_by_key` into agent `read_chunks_by_key_inner`; key carrier switched from row-major `logical_value_t` to columnar `data_chunk_t` (mirrors `scan_by_keys`); all prod + test callers updated.
- Added multi-key `read_chunks_by_keys` and converted the per-key read loops in `operator_resolve_constraint` and `operator_dynamic_cascade_delete`.
- Deleted prod-dead `resolve_table`/`resolve_type`/`resolve_function` (+ `resolve_type_sync`) and their dead result structs; disk tests use a `catalog_probe` helper over `read_chunks_by_key`.

**Track C — dead-code sweep (disk-service API)**
- Removed the singular MVCC trio, `total_rows_sync`/`checkpoint_wal_id_sync` chain, `role()`/`pool_idx()`/`make_type()`, `lv_i32/i64/bool`, `load_system_tables_sync`, `oid_gen()`, agent `direct_append_sync`, `storage_column_names_inner`, `drop_column_inner`, `table_storage_t::table() const`, and the redundant 4-arg `direct_append_sync`/`direct_delete_sync` overloads.

**Track D — batch conversions**
- `drop_storage_many` + `mark_storage_dropped_many` (per-agent partition); FK cascade SET NULL/DEFAULT collapsed to one fetch + one update; commit publish 4 round-trips → 2.
- `create_storage`/`_with_columns`/`_disk` converted to mailbox agent handlers (agent builds the SFBM on its own thread); `checkpoint_all` folds `has_in_memory` into `checkpoint_inner`'s return.

**Index**
- `flush_all_indexes` converted to a `force_flush` mailbox fan-out; removed the runtime `agent->force_flush_sync()`/`is_dropped()` cross-actor sync calls.

**Infra / tests**
- Extracted `services/disk/inline_scan.hpp`; added agent `scan_batched_local`.
- Added targeted tests (catalog WAL ordering, WAL-disabled append, `read_chunks_by_keys` parity, `drop_storage_many`, `mark_storage_dropped_many`, DROP INDEX delete fold).

### DIFF
--- a/components/physical_plan/operators/operator_data.cpp
+++ b/components/physical_plan/operators/operator_data.cpp
@@ -121,4 +121,49 @@ namespace components::operators {
         return out;
     }
 
+    vector::data_chunk_t make_key_chunk(std::pmr::memory_resource* resource,
+                                        std::pmr::vector<types::logical_value_t> values) {
+        // Column j carries the value's own complex_logical_type, so the cell is written
+        // without a cast. Capacity is at least 1 because set_cardinality(1) below requires
+        // capacity_ >= 1 even when there are no key columns.
+        std::pmr::vector<types::complex_logical_type> types(resource);
+        types.reserve(values.size());
+        for (const auto& v : values) {
+            types.emplace_back(v.type());
+        }
+        vector::data_chunk_t chunk(resource, types, 1);
+        for (std::size_t j = 0; j < values.size(); ++j) {
+            // set_value already leaves the cell invalid for a null logical_value_t.
+            chunk.set_value(static_cast<uint64_t>(j), 0, values[j]);
+        }
+        chunk.set_cardinality(1);
+        return chunk;
+    }
+
+    vector::data_chunk_t make_keys_chunk(std::pmr::memory_resource* resource,
+                                         const std::pmr::vector<std::pmr::vector<types::logical_value_t>>& rows) {
+        // Column types are taken from the first key-tuple; every other tuple must be
+        // positionally aligned to the same key columns (same arity, same per-column type).
+        // Capacity is at least 1 because set_cardinality requires capacity_ >= 1 even when
+        // there are no rows.
+        std::pmr::vector<types::complex_logical_type> types(resource);
+        if (!rows.empty()) {
+            types.reserve(rows.front().size());
+            for (const auto& v : rows.front()) {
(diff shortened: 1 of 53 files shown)


components/physical_plan/operators/operator_data.cpp
```diff
@@ -121,4 +121,49 @@ namespace components::operators {
...
+        for (std::size_t i = 0; i < nrows; ++i) {
+            const auto& row = rows[i];
+            for (std::size_t j = 0; j < row.size(); ++j) {
+                // set_value already leaves the cell invalid for a null logical_value_t.
+                chunk.set_value(static_cast<uint64_t>(j), static_cast<uint64_t>(i), row[j]);
+            }
+        }
+        chunk.set_cardinality(static_cast<uint64_t>(nrows));
+        return chunk;
+    }
```
OrangeSmokingJacket, line 167: Populating data_chunk_t through logical_value_t is extreamly slow, and those functions are used a lot, mainly because there is no clean way to set string value. I will look into that

PR #518, opened 2026-06-13
hash-index-v1.2-fix

AnnaBekarevitch:


services/index/disk_hash_table.cpp
```diff
@@ -24,6 +26,21 @@ namespace services::index {
             return h;
         }
 
+        uint32_t generate_hash_seed() {
+#ifdef DEV_MODE
+            if (const char* v = std::getenv("OTTERBRIX_DISK_HASH_SEED"); v != nullptr && *v != '\0') {
+                return static_cast<uint32_t>(std::strtoul(v, nullptr, 0));
+            }
+#endif
+            std::random_device rd;
```
OrangeSmokingJacket, line 35: std::random_device is a heavy object. If generate_hash_seed() is called only a couple of times, it should be fine, otherwise It would make sence to make 'rd' static or thread_local

OrangeSmokingJacket (commented):
Looks like a solid fix. But since bug was missed by our CI/CD pipeline, and only surfaced in benchmarks, we should do something about it. Can you add some tests for it?

services/index/disk_hash_table.cpp
AnnaBekarevitch: it is called only on index creation

PR #519, opened 2026-06-14
PAX-format realisation

IWantToWorkInGoogle:
## Summary

Introduces a **PAX-like (Partition Attributes Across) on-disk storage format** for
disk-backed tables — a hybrid between row and pure-columnar layouts. Within a row
group, values are stored column-by-column (enabling columnar scan, zone-map pruning
and better compression) while keeping each row group self-contained on disk. It lives
alongside the existing columnar storage and is selectable per table.

## What's included

**Storage format**
- PAX on-disk format with encodings for fixed-width, string, and nested
  (struct / list / array) columns.
- Full checkpoint → load round-trip; right-sized validity so cold reopen/load scales
  to large tables.
- Per-table storage-format selection, observable via `pg_class.relstorageformat`
  (`disk_auto` / `disk_pax` / `disk_columnar`).

**Scan & query**
- Zone-map (min/max) pruning for range and conjunction predicates pushed into the scan.
- Opt-in parallel scan (`OTTERBRIX_PARALLEL_SCAN`).
- PAX-aware scan / index planning and operators.
- Uncorrelated subquery support (unblocks TPC-H q11 / q18 at the SQL layer).

**Durability & robustness**
- Bounds-checked decode/load paths against corrupt or truncated data.
- Per-block CRC32c verification and a checksummed DB header.
- Fail-closed checkpoint writes (dropped block writes are no longer silently swallowed).
- Fixes for validity-mask corruption across checkpoint/reopen and append-after-reopen.

**Foundation**
- Generic `type_spec` type system underpinning the column encodings; string-heap support.
- Disk service: PAX storage adapter, agent pool, and scan/checkpoint path.

## Testing
- PAX checkpoint/load round-trip and storage round-trip tests, plus a **differential
  fuzzer** that verifies a cold-reopened index/table matches an in-memory oracle
  (including delete/update paths).
- Full local test suite passes.

## Commit layout (10 logical commits)
1. `types` — generic `type_spec` + string-heap foundation
2. `table` — PAX on-disk storage format (fixed/string/nested encodings)
3. `table` — column-data + row-group checkpoint/load engine
4. `table` — PAX checkpoint/load + round-trip tests
5. `vector` — scan/gather operations for PAX
6. `disk` — PAX storage adapter, agent pool, scan/checkpoint path
7. `planner` — PAX-aware scan/index planning + operators
8. `sql` — uncorrelated subquery support (TPC-H q11/q18)
9. `catalog/dispatcher/wal` — PAX DDL metadata + storage-format wiring
10. `integration + build` — end-to-end PAX tests + cmake wiring

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ endif()
 # We do not use -Werror flag with sanitizers for the following reasons:
 # 1) Sanitizers force us to build absl locally, which does contains warnings
 # 2) Incomplete support of macOS makes development too cumbersome with -Werror enabled
-set(USE_Werror "$<NOT:$<OR:$<BOOL:${APPLE}>,$<BOOL:${ENABLE_ASAN}>,$<BOOL:${ENABLE_UBSAN}>>>")
+set(USE_Werror "$<NOT:$<OR:$<BOOL:${APPLE}>,$<BOOL:${ENABLE_ASAN}>,$<BOOL:${ENABLE_TSAN}>,$<BOOL:${ENABLE_UBSAN}>>>")
 
 if ((CMAKE_CXX_COMPILER_ID MATCHES "GNU") OR (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
     add_compile_options(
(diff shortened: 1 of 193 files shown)


morgangt:
Hello! 
Is this PR still active? I'm asking because I'm working on
something that touches the same scan path and I'd like to avoid duplicating work
or creating merge conflicts.

My idea is a runtime Bloom filter for hash joins (sideways information passing):

- Phase 1 (done locally): a split-block Bloom filter built over the build-side key
  hashes inside operator_hash_join_t. The probe side tests it before the
  equal_range lookup, so miss-heavy probes (typical for star-schema joins) skip
  the hash-table walk at the cost of one cache-line read. No semantic changes —
  the filter is keyed on the same data_chunk_t::hash values as the index, so
  false negatives are impossible and a false positive just falls through to the
  existing verify path.

- Phase 2 (planned): pushing that filter below the join into the probe-side scan.
  The executor materializes fragments in build → probe → join order, so the probe
  scan can lazily build the filter from the already-materialized build output and
  ship it to the disk agent inside table_filter_t. Applying it in
  row_group_t::filter_indexing (before the residual predicate graph) would shrink
  the selection vector before non-key columns are materialized, reusing the
  existing skip / late-materialization machinery.

This is where the overlap comes in: this PR introduces zone-map (min/max) pruning
and its own scan/planning path for PAX tables, while my phase 2 extends
table_filter_t / filter_indexing in the current columnar path. A few questions:

1. Is the PAX work still moving forward, and what's the rough timeline?
2. Is zone-map pruning intended to stay PAX-only, or is there a plan to enable it
   for the existing columnar storage too (check_zonemap_segments is currently
   stubbed out there)?
3. If both land, would you prefer the runtime join filter to hook into a shared
   filter representation so PAX and the current format can consume it uniformly?

Happy to align the design with your plans — I'd rather build on top of this PR's
scan changes than race it. 

Thanks!

PR #520, opened 2026-06-14
refactor(python): ingest pandas/numpy via Arrow C-stream, drop the bespoke scanner

kotbegemot:
4 commits on top of `main` (which already contains the pandas/numpy/polars integration).

## What
pandas >= 2.2 exposes `__arrow_c_stream__` for every DataFrame (`pa.Table.from_pandas`),
so the ~2.8k-line bespoke numpy/pandas column scanner is redundant. Route all pandas
DataFrame and raw-numpy ingestion through the existing core Arrow path and delete the
scanner; plus the core fixes/migration that path required.

## Commits
1. **fix(vector/arrow,types)** — core Arrow-consumer fixes (latent/untested before):
   validity-mask NULL allocation (columns with NULLs segfaulted); MAP physical-type +
   struct<key,value> entry type via `child_type` + `value`/`set_value` reconciled to the
   list-of-struct representation; `create_map` threads a `memory_resource*` for the entry
   struct (no `get_default_resource`).
2. **refactor(python)** — delete `integration/python/{pandas,numpy}/`; add the
   `pandas_arrow_prepare` shim (dedup, dict->MAP, lenient string, decategorize); wrap raw
   numpy in a DataFrame (preserves NaN->NULL); relocate the input-kind enum; pin
   `pandas>=2.2,<4`; bump `requires-python>=3.9`; add dtype tests.
3. **refactor(vector/arrow)** — migrate the Arrow ingest conversion chain from C++
   exceptions to `core::result_wrapper_t` / `core::error_t` (~32 throw sites). Entry points
   (`schema_from_arrow`, `data_chunk_from_arrow`, type parsing, `get_next_chunk`) return
   `result_wrapper_t<T>`; the recursive column scanners return `[[nodiscard]] core::error_t`
   so `-Werror` catches any unpropagated error. Error messages use a threaded
   `memory_resource`. The pybind boundary unwraps and surfaces errors as exceptions (the
   established convention); the core conversion is now exception-free. Export/appender path
   left unchanged.

## Behavior
NaN->NULL, nullable Int64, datetime/NaT, dict->MAP, dict->struct, mixed-object->STRING
(lenient), duplicate-column dedup, numpy 1D/2D/list/dict, categoricals.

## pandas 3
Allowed (`<4`); the `__arrow_c_stream__` producer contract was verified identical to
pandas 2.x. pandas 3 requires Python >= 3.11, so the CI 3.11 runner exercises it.

## Tests
- pytest dataframe suite: 33/33
- C++ ctest: 880/880

## Dictionary / categorical decode (commit 4)
The direct pyarrow `dictionary` / polars categorical decode path crashed (use-after-move +
ownership double-free). Both are fixed, so null-free dictionary/categorical columns now decode
correctly. Columns containing null values are reported via `core::error_t` (the null-index
decode is a separate pre-existing bug) rather than crashing or returning wrong data. The pandas
path is unaffected — categoricals are decategorized before export.


integration/python/scan/pandas_arrow_prepare.cpp
```diff
@@ -0,0 +1,143 @@
...
+                value_type = pa.attr("string")();
+            }
+            py::object map_type = pa.attr("map_")(key_type, value_type);
+            return pa.attr("array")(rows, py::arg("type") = map_type);
+        }
+
+    } // namespace
+
+    py::object prepare_dataframe_for_arrow(const py::object &df_in) {
+        py::object df = df_in.attr("copy")(py::arg("deep") = false);
```
prawwtocol, line 80: nit:


we used to ignore pandas indexes

( `df = pd.DataFrame({"x": [10, 20, 30]}, index=["alice", "bob", "carol"])`)

see here: https://github.com/otterbrix/otterbrix/blob/main/integration/python/pandas/bind.cpp#L43

now it seems like the index would be added as a column (https://arrow.apache.org/docs/python/pandas.html#handling-pandas-indexes), do we need this?

prawwtocol (approved):

PR #525, opened 2026-06-20
fix(table): disk-backed buffer cache + eviction guard to fix HTAP scan segfault

kotbegemot:
## Problem

Under memory pressure, large-table (HTAP) scans segfaulted. Root cause: the
buffer pool was used as a **store, not a cache**. Filled column segments had
no backing store, so when the pool evicted such a block, a later re-pin
dereferenced a null buffer → `EXC_BAD_ACCESS`.

## Fix

Turn the buffer pool into a real disk-backed cache:

- `block_handle` distinguishes **reloadable** blocks
  (`block_id < MAXIMUM_BLOCK`, a disk copy exists and `load()` can re-read
  it) from **managed in-memory** blocks (`load()` returns `{}`).
- **Eviction guard**: `can_unload()` is gated on `is_reloadable()`, so a
  block with no backing store is never unloaded — a re-pin can no longer
  hit a null buffer.
- **Write-through**: a filled segment is written to the data file and
  swapped for a disk-backed, evictable+reloadable segment
  (`transition_segment_to_disk` + `segment_tree::replace_segment_at_index`),
  so memory stays bounded at any table size.
- **No throwing on the agent thread**: `load()` / `buffer_pool::set_limit()`
  and the scan/update/append chains forward
  `data_corruption` / `io_error` / `out_of_memory` / `write_conflict`
  through `result_wrapper` instead of throwing. The `agent_disk` /
  `manager_disk` replies carry the error, so the executor surfaces a clean
  error cursor.

Disk manager bootstrap/IO/resolve and the table storage adapter are
updated for the disk-backed segment lifecycle.

## Scope

A table that exceeds its pool now returns a clean `out_of_memory` error
cursor instead of segfaulting, and large scans that fit on disk complete
with bounded RSS. **Bounded-memory streaming execution** (push-based scan
pipeline) is separate follow-up work, not part of this PR.

## Tests

- New: `test_disk_backed_scan`, `test_eviction_guard`,
  `test_arithmetic_empty`.
- Extended: checkpoint load, block manager, persistence, MVCC,
  transaction manager, and production-scenario coverage.

### DIFF
--- a/components/vector/arithmetic.cpp
+++ b/components/vector/arithmetic.cpp
@@ -1121,6 +1121,14 @@ namespace components::vector {
                                        const vector_t& left,
                                        const vector_t& right,
                                        uint64_t count) {
+        // Empty input: produce an empty result without dereferencing operand
+        // vectors. On a 0-row chunk the operands may not be resolvable at all
+        // (e.g. a degenerate 0-column batch chunk yields out-of-bounds operand
+        // pointers), so reading left.type()/right.type() here would deref a
+        // dangling vector. A 0-row arithmetic result carries no values.
+        if (count == 0) {
+            return vector_t(resource, types::complex_logical_type(types::logical_type::DOUBLE), 0);
+        }
         if (types::is_duration(left.type().type()) || types::is_duration(right.type().type())) {
             return compute_temporal_binary(resource, op, left, right, count);
         }
@@ -1159,6 +1167,12 @@ namespace components::vector {
                                               const vector_t& vec,
                                               const types::logical_value_t& scalar,
                                               uint64_t count) {
+        // Empty input: see compute_binary_arithmetic. The vec operand may be a
+        // dangling/out-of-bounds reference on a 0-row chunk, so do not read its
+        // type here.
+        if (count == 0) {
+            return vector_t(resource, types::complex_logical_type(types::logical_type::DOUBLE), 0);
+        }
         if (types::is_duration(vec.type().type()) || types::is_duration(scalar.type().type())) {
             return compute_temporal_vec_scalar(resource, op, vec, scalar, count);
         }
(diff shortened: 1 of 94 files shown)


components/vector/arithmetic.cpp
```diff
@@ -1121,6 +1121,14 @@ namespace components::vector {
...
                                        const vector_t& right,
                                        uint64_t count) {
+        // Empty input: produce an empty result without dereferencing operand
+        // vectors. On a 0-row chunk the operands may not be resolvable at all
+        // (e.g. a degenerate 0-column batch chunk yields out-of-bounds operand
+        // pointers), so reading left.type()/right.type() here would deref a
+        // dangling vector. A 0-row arithmetic result carries no values.
+        if (count == 0) {
+            return vector_t(resource, types::complex_logical_type(types::logical_type::DOUBLE), 0);
+        }
```
OrangeSmokingJacket, line 1131: Why did you pick 'double' as default type? Also, this does look like it shouldn't be called with 0 count at all

integration/cpp/base_spaces.cpp
```diff
@@ -160,6 +160,18 @@ namespace otterbrix {
...
+            // storage back. Reconstruct the (empty) storage shell for any alive
+            // user table the catalog knows about but whose storage is still
+            // missing, from its pg_attribute columns. Without this, a reopened
+            // session's CREATE TABLE IF NOT EXISTS finds the table "exists" and
+            // skips creating storage, resolve_table returns a schema, but every
+            // INSERT silently no-ops (storage_append returns 0,0) and scans see
+            // nothing. Runs after load_user_table_storages_sync so on-disk tables
+            // (already loaded) are skipped, and before WAL replay so replayed
+            // INSERTs land in the rehydrated storage.
+            disk_ptr->rehydrate_in_memory_user_storages_sync();
```
OrangeSmokingJacket, line 174: Aren't in-memory tables just suppose to vanish on reload?

services/disk/manager_disk_bootstrap.cpp
```diff
@@ -326,6 +348,17 @@ namespace services::disk {
...
+            // it is a UINTEGER (oid_col()). Some system tables (e.g. pg_settings)
+            // key on a STRING column 0 (`name`); reading that as a uint32 OID
+            // yields garbage that would poison oid_gen_ with a huge, non-
+            // deterministic high_water — every fresh CREATE TABLE then mints a
+            // wild OID, and on reopen the persisted (garbage) catalog OID no longer
+            // matches the storage the agent loaded, so user-table appends silently
+            // no-op. Only scan column 0 when it is an OID column.
+            if (table.columns()[0].type().type() != components::types::logical_type::UINTEGER) {
+                continue;
+            }
```
OrangeSmokingJacket, line 361: Shouldn't it be an error, if fist column is not uint32?

PR #528, opened 2026-06-23
feat(htap): push-based streaming executor + storage segment packing

kotbegemot:
## Summary

Closes the HTAP large-table scan segfault and makes disk-backed SELECT execution
bounded, then attacks the disk over-allocation that drove committed memory up. Four
commits, building on the disk-backed buffer cache:

1. **`fix(table)`: disk-backed buffer cache + eviction guard** (STEP 1/2) — the scan
   segfault becomes a clean error cursor; large scans read from disk in bounded batches.
2. **`refactor(table)`: `[[nodiscard]]` + precise error codes** — error-returning storage
   APIs surface failures as values, never throws across the mailbox.
3. **`fix(table)`: satisfy CI `-Werror`** after the error-handling pass.
4. **`feat(htap)`: push-based streaming executor + storage segment packing** (this PR's
   main commit) — described below.

## STEP 3 — bounded-memory SELECT execution

- Read-path operators converted to a **push-based pipeline** (source / streaming / sink
  roles in `operator.hpp`; `execute_pipeline()` drive loop). `full_scan` / `transfer_scan`
  are fetch-next sources; `select` streams; `hash_join`, `aggregate`, `group`, `sort`
  accumulate bounded sink state and emit at `finalize()`. Peak memory = one batch + active
  sink state.
- **Fetch-next index-resume cursor** in `agent_disk` (`active_scans_`, position-only): each
  `storage_fetch_next_batch` re-seeks by index, reads one batch, advances the stored
  position, and holds **zero pins** across the round-trip, so the pool evicts freely.
- **R1**: `logical_value_t` join/group keys replaced by a **typed hash+verify** index.
- **R6**: the `operator_group` `calc_aggregate_values_fallback` is deleted (single path).

## Storage segment packing ("B2") — over-allocation fix

- Write-through segment persistence now routes through `partial_block_manager_t`, so
  multiple small column segments **share one disk block at distinct offsets** instead of
  each grabbing a dedicated 256 KiB block. Wide-table microbench: **480 → 15 blocks (32×)**.
  Flush-before-evict is enforced at every `transition_segment_to_disk` caller boundary.
- Fixes two latent bugs the packing exposes: the compaction reclaim leaked packed block
  ids (unbounded growth), and the checkpoint UNCOMPRESSED path ignored
  `column_segment::block_offset()` (could copy a neighbour's bytes — data corruption).

## Concurrent-checkpoint corruption fix ("A")

- Guard `block_handle_t` teardown under its lock and pair `mark_as_free` with
  `unregister_block` in the compaction reclaim to kill an ABA on the shared pool free-list
  (previously SIGSEGV / SIGABRT under concurrent checkpoints).

## Memory footprint (BUG B) — partially closed, decision documented

`docs/bug-b-footprint-decision.md` records the post-packing measurement: committed memory
at 300K×16 dropped from ~2007 MB to **282 MB** (GB-scale ratchet gone), but still plateaus
at the working-set peak (the shared pool caches it; blocks are ~26 % full at
`row_group_size = 1024`). The residual is an **efficiency** decision — accept the plateau
(A), densify blocks via larger row groups (B1), or add a trimming pool (C) — not a
correctness blocker.

## Tests

- Full suite **942/942** (Debug).
- New RED→GREEN guards: bounded-execution streaming, eviction-reload (flush-before-evict),
  over-allocation bound, lost-wakeup real-path/streaming, and correctness-bug regressions.

## Not in this PR (follow-ups)

- **STEP 3b** — streaming DML (insert/update/delete/FK); DML operators still materialize
  for atomicity (role `none`) by design.
- **STEP 4** — single-node spillable sinks (for intermediates that exceed memory).
- **B1 / C** — the BUG B footprint follow-ups above.


components/physical_plan/operators/operator_delete.cpp
```diff
@@ -40,171 +39,260 @@ namespace components::operators {
...
+        // (the scan already pushed the WHERE), so create_all_true_predicate matches
+        // every scan row; a non-null expression_ (legacy callers) is honored too.
+        auto predicate = expression_ ? predicates::create_predicate(resource_,
+                                                                    pipeline_context->function_registry,
+                                                                    expression_,
+                                                                    types,
+                                                                    types,
+                                                                    &pipeline_context->parameters,
+                                                                    pipeline_context->session_tz)
+                                     : predicates::create_all_true_predicate(resource_);
```
OrangeSmokingJacket, line 62: Are types allowed to change during the stream? For constant schema we can build predicate once, but for computed schema it is safer to do it that way

PR #534, opened 2026-07-04
bounded DML sinks, agent-side aggregate pushdown, UNIQUE/PK + CHECK enforcement

kotbegemot:
Step 3b: memory-bounded DML sinks, single-owner aggregate pushdown, and UNIQUE/PK + CHECK enforcement — review-hardened (a 10-finding multi-agent review pass, every fix landed red-first with its own regression test).

## DML bounding (3b-B)

- `insert` / `update` / `delete` become **bounded streaming sinks**: the executor drives incremental mid-pump flushes whenever `buffered_rows()` crosses `config.execution.dml_flush_row_threshold` (default `0` = single post-pump flush, behavior-preserving). Common per-flush bookkeeping lives in `dml_util.hpp` (no base class, no `std::function`).
- The mid-flush gate is wired into **all three** pump branches, including the producing sourceless bottom (recursive CTE feeding a DML sink) — `INSERT INTO t (n) WITH RECURSIVE ... SELECT` honors the bound.
- **Failed-statement atomicity**: an INSERT RETURNING projection error carries the already-appended range with the error so the abort tail reverts the physical append; UPDATE RETURNING projects **before** any storage mutation (clean failure: zero writes, no WAL record, nothing to revert — pinned by the DEV_MODE `update_storage_update_sends()` counter).

## Aggregate pushdown (SEAM B)

- The optimizer rule stamps single-owned-table `SUM/COUNT/MIN/MAX/AVG` aggregates (no HAVING / DISTINCT / UDF); physgen lowers them to `operator_group_merge` over `pushed_reduce_scan`.
- The scan ships a POD `pushed_aggregate_spec_t` on a **dedicated `storage_reduce` protocol leg**: the owning agent rebuilds the existing `operator_group` from the POD, reduces its slice send-free, and replies **all final rows in one reply** (bounded by `#groups`, no cursor). `storage_fetch_next_batch` stays a pure scan — no modal spec parameter.
- `operator_group_merge` is the coordinator-side aggregate terminal: identity passthrough under the **single-owner invariant** (one agent owns the whole table — pinned in `disk_contract.hpp`), owner of the empty-input scalar row (`COUNT=0` / `NULL`), and the socket where sharded slices would plug a real partial-state kernel merge without reshaping the plan.
- Traffic contract: a scalar aggregate crosses the mailbox as exactly 1 row, a GROUP BY as one row per group (`test_pushdown_traffic`).

## Constraint enforcement

- **`operator_unique_constraint` (3b-D)**: UNIQUE/PK via within-batch typed hash+verify plus an existing-row `scan_by_keys` layer. DEFAULT-backed columns absent from a *name-addressed* write-set are materialized as constant key columns (an `INSERT (a)` omitting a `UNIQUE ... DEFAULT 5` column still collides); `cells_equal` floats compare exactly (`NaN==NaN`) to keep the hash/verify contract.
- **`operator_check_constraint`**: CHECK predicates and the NOT NULL pass evaluate absent columns against decoded column DEFAULTs (enrich decodes `attdefspec` and stamps the DML nodes); PRIMARY KEY columns are merged into the NOT-NULL set (ALTER ADD PRIMARY KEY never back-fills `attnotnull`). Positional / hand-built inserts keep the legacy pass-through (`write_set_named` gate).
- **FK (3b-C)**: single-pass hash semi-join on the agent replaces the per-key eq-scans.

## Batched CREATE INDEX backfill (3b-E)

Streams the table through the fetch-next cursor in bounded batches; index entries are stamped with each batch's **true (gapped) physical `row_ids`**, one contiguous run per `insert_rows` call — MVCC-skipped (deleted) rows can no longer shift the index→row mapping.

## Review hardening

Ten confirmed findings fixed red-first, each with its own regression test: backfill row-id misalignment after DELETE; UNIQUE skip on defaulted columns (within-batch + existing-row); ALTER ADD PK admitting NULL keys; INSERT/UPDATE RETURNING atomicity; re-driven pushed scan losing its spec (now structurally impossible — the spec ships as a copy per drive); pushed aggregate over a missing/record-only slice returning zero rows instead of `COUNT=0`; the NaN hash/equality contract; the missing mid-flush gate on the producing pump branch; a CHECK `IS NOT NULL` false-reject of DEFAULT-backed columns.

## Tests

- integration: **336/336** (55,067 assertions), incl. new `test_bounded_dml_flush`, `test_aggregate_pushdown_e2e`, `test_pushdown_traffic`, `test_unique_constraint_{e2e,operator}`, `test_stacked_constraints`, `test_create_index_backfill_batched`
- disk: **130/130** (incl. `test_pushdown_reduce` protocol/merge units, `test_fk_scan_by_keys_semijoin`)
- planner / dispatcher / sql / table / wal / index suites green


components/planner/optimizer/rules/pushdown_aggregate.cpp
```diff
@@ -0,0 +1,247 @@
...
+        // insensitively: the SQL parser lowercases unquoted function names, but
+        // guard against a quoted/upper spelling.
+        bool is_mergeable_kind(std::string_view function_name) noexcept {
+            std::string lowered(function_name);
+            std::transform(lowered.begin(), lowered.end(), lowered.begin(), [](unsigned char c) {
+                return static_cast<char>(std::tolower(c));
+            });
+            static constexpr std::array<std::string_view, 5> mergeable{"sum", "count", "min", "max", "avg"};
+            return std::find(mergeable.begin(), mergeable.end(), lowered) != mergeable.end();
+        }
```
OrangeSmokingJacket: Functions are registered at runtime by the user.
1) It will be easy to forget to update it, when new built-in function is added
2) It does not accout for user defined aggregate function
3) What is the difference between usual fragment merge and per-agent merge?

components/physical_plan/operators/operator_unique_constraint.cpp
```diff
@@ -0,0 +1,273 @@
...
+                    col_names.emplace_back(n);
+                }
+
+                auto [_, fut] = actor_zeta::send(ctx->disk_address,
+                                                 &services::disk::manager_disk_t::scan_by_keys,
+                                                 exec_ctx,
+                                                 table_oid_,
+                                                 std::move(col_names),
+                                                 std::move(keys));
+                auto matches = co_await std::move(fut);
```
OrangeSmokingJacket, line 309: This pattern is still surprisingly common in our code. N number of sends and awaits, instead of doing it all at ones

components/physical_plan/operators/scan/pushed_reduce_scan.cpp
```diff
@@ -0,0 +1,100 @@
...
+            // non-all_false compare, so transform_predicate never trips its assert.
+            std::unique_ptr<table::table_filter_t> filter;
+            if (expression_ && expression_->type() != expressions::compare_type::all_true) {
+                auto [_t, tf] = actor_zeta::send(ctx->disk_address,
+                                                 &services::disk::manager_disk_t::storage_types,
+                                                 ctx->session,
+                                                 table_oid_);
+                auto types = co_await std::move(tf);
+                auto filter_result =
+                    transform_predicate(resource_, expression_, types, &ctx->parameters, ctx->session_tz);
```
OrangeSmokingJacket: I think that predicate should only be built once, not per each source_next call. But I can see how that might be a problem with computed schema

OrangeSmokingJacket (approved):

PR #538, opened 2026-07-08
Upgrade Catch2 v2.13.7 to v3.15.1

### LINKED ISSUE #273: upgrade Catch2 v3

cleveralbatraoz:
Fixes #273

Upgrades Catch2 from v2.13.7 to v3.15.1. Catch2 v3 is a compiled library with split headers and no `CATCH_CONFIG_MAIN`, so beyond the version bump (`conanfile.py`, `find_package(Catch2 3 REQUIRED)`) the migration entails:

- Replace the monolithic `<catch2/catch.hpp>` include in every test file with the specific v3 headers each file actually needs: `catch_test_macros.hpp` everywhere, plus `catch_template_test_macros.hpp`, `catch_approx.hpp` and `generators/catch_generators.hpp` only where those features are used.
- Delete the 22 boilerplate `CATCH_CONFIG_MAIN` `main.cpp` files and link `Catch2::Catch2WithMain` instead. Targets whose main file holds real test cases (`example/cpp`, `integration/c/test`, `benchmark/runner/test`) keep their sources and just switch the link. `profile_arithmetic` has its own `main()` and keeps plain `Catch2::Catch2`.
- Qualify `Approx` as `Catch::Approx` (v3 no longer exports it unqualified).
- Terminate `INFO(...)` statements before block scopes (~1000 sites; see the styling note below).
- Add std includes that v2's `catch.hpp` used to provide transitively:
  - `<random>` in three test files (unconditionally — in `test_b_plus_tree.cpp` it must sit outside the `#if defined(__linux__)` include block or macOS loses it);
  - `<memory>` ahead of every `<actor-zeta/spawn.hpp>` include: `spawn.hpp` uses `std::unique_ptr` without including `<memory>`, and gcc 11's libstdc++ doesn't pull it in transitively, so the disk tests only compiled because `catch.hpp` included it first. The real fix belongs upstream in actor-zeta; the explicit includes carry a comment explaining why.
- Parenthesize the `file_time_type` comparison in `test_system_table_bootstrap.cpp`: Catch2 v3 stringifies assertion operands by default (chrono support was opt-in in v2), and Apple's `std::filesystem` clock uses an `__int128` tick representation that `ostream` cannot stream — `REQUIRE(a == b)` on file times breaks the macOS build inside Catch2's own headers. `REQUIRE((a == b))` keeps the operands opaque.

## Styling note (why some braces moved)

The codebase used `INFO("x") { ... }` as pseudo-section labels. That only compiled because **v2's `INFO` macro expansion carried a trailing semicolon**; v3 dropped it, so the semicolon is now explicit and the `{...}` becomes a standalone compound statement. Brace placement of these blocks follows the project's own formatter (clang-format-11 per `scripts/README.md`, with the repo's `.clang-format`, which is untouched by this PR): multi-statement blocks get `{` on its own line, trivially short blocks collapse to `{ stmt; }`. All migrated lines were formatted with an actual clang-format 11.1.0 binary. Message lifetime and scoping of the `INFO`s are unchanged.

## Verification

- ctest registration is **unchanged**: 1093 tests with byte-identical names before and after (verified against a clean Catch2 v2 build of `main`); all 1093 pass locally.
- Full rebuild with gcc 14.2 (Release, `DEV_MODE=ON`, `EXAMPLE=ON`); all 21 component/service test binaries pass.
- Syntax-swept all 478 TUs with a real gcc 11.4 toolchain (reproduces CI's compiler): zero failures.
- Differential clang 17 + libc++ sweep against `main`: zero migration-caused failures.
- The example binary behaves identically to `main` (including its one pre-existing `example::sql::group_by` failure, unrelated to this change and never run in CI).


cleveralbatraoz:
one moment, i'll fix it

OrangeSmokingJacket (approved):

kotbegemot (approved):

PR #545, opened 2026-07-12
fix(sql): bare subquery in predicate position crashed the server

cleveralbatraoz:
A bare subquery in predicate position segfaults the server (Release build), verified against current `main`:

```sql
SELECT * FROM t WHERE (SELECT x FROM t);   -- segfault
```

## Root cause

`transform_sublink_expr` implements only `EXISTS_SUBLINK`, `ANY_SUBLINK` and `ALL_SUBLINK`. The six other `SubLink` kinds (`NOT_EXISTS`, `ROWCOMPARE`, `EXPR`, `ARRAY`, `CTE`, `INITPLAN_FUNC`) all `break` out of the switch into a terminal `assert(false)` — which NDEBUG erases, so execution **falls off the end of a non-void function**: undefined behavior, observed as a crash on the `EXPR_SUBLINK` reproduction above.

## Fix

The fall-through now sets a clean transformer error ("unsupported subquery expression in this context") and returns null, matching the error convention used elsewhere in the same function. The handled EXISTS / ANY / ALL / `IN (SELECT …)` paths are untouched.

## Testing

- New case in `components::sql::errors` asserting the clean error message — it crashed the test binary before the fix (the transform executed UB).
- The full `components::sql` suite (44 tests) passes.

### DIFF
--- a/components/sql/test/test_errors.cpp
+++ b/components/sql/test/test_errors.cpp
@@ -159,4 +159,12 @@ TEST_CASE("components::sql::errors") {
     TEST_PARSER_ERROR("CREATE INDEX base ON TEST_DATABASE.TEST_COLLECTION;", R"_(syntax error at or near ";")_");
 
     TEST_TRANSFORMER_ERROR("DROP INDEX TEST_DATABASE.TEST_COLLECTION;", R"_(incorrect drop: arguments size)_");
+
+    // A bare subquery in predicate position (EXPR_SUBLINK) — and the other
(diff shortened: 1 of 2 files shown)


cleveralbatraoz:
Filed #559 to track the follow-up: this PR converts the crash into a clean "unsupported subquery expression in this context" error, so the underlying feature gap (scalar/`EXPR` subqueries in value position, plus `ARRAY(SELECT …)` and row-comparison SubLinks) is now recorded there and won't be lost.

PR #565, opened 2026-07-14
fix(table): NULL must not satisfy a pushed-down value comparison (SQL three-valued logic)

cleveralbatraoz:
## The bug

A `NULL` silently matches every value comparison that `0` satisfies, and `DELETE`/`UPDATE` act on those rows. Plain table, declared columns, one honest NULL:

```sql
CREATE TABLE hr.emp (id INT, name TEXT, bonus BIGINT);
INSERT INTO hr.emp VALUES (1, 'ann', 500);   -- got a bonus
INSERT INTO hr.emp VALUES (2, 'bob', NULL);  -- bonus not yet decided
INSERT INTO hr.emp VALUES (3, 'cid', 0);     -- got nothing

SELECT name FROM hr.emp WHERE bonus IS NULL;  -- bob        (correct: the NULL is really there)
SELECT name FROM hr.emp WHERE bonus = 0;      -- bob, cid   WRONG, expected: cid
SELECT name FROM hr.emp WHERE bonus < 1;      -- bob, cid   WRONG, expected: cid
SELECT name FROM hr.emp WHERE bonus >= 0;     -- ann,bob,cid WRONG, expected: ann, cid

DELETE FROM hr.emp WHERE bonus = 0;           -- "clear out everyone who got no bonus"
SELECT name FROM hr.emp;                      -- ann        WRONG: bob is gone
```

Bob's bonus was undecided, not zero. No error, no warning — the statement reported success and deleted him.

## Root cause

A pure comparison is planned as a **pushdown filter into the storage scan** (`create_plan_match.cpp:107-136`), not as `operator_match`/`simple_predicate`. The scan's fast path (`column_data_t::check_predicate`) pins the raw block and compares the **data buffer**, never consulting the validity mask. Validity lives in a physically separate sibling column (`standard_column_data_t`), so a NULL row's payload bytes are a meaningless `0` — hence `0 == 0` matches.

The same function guards this in its STRUCT/ARRAY/LIST branch and its RLE/DICTIONARY branch, and omits it in the plain fixed-size/string branch — the one every ordinary column takes. `simple_predicate`, `IS NULL`, and the projection path all handle NULL correctly; only the pushed-down scan filter departed from SQL semantics.

## Why the fix is not just a validity guard

Adding the missing guard fixes the comparisons and then **breaks `NOT`**:

```sql
SELECT id FROM t WHERE NOT (x = 0);   -- guard alone: returns the NULL row. Expected: it must not.
```

The filter tree was two-valued (`bool`), so `union_not` simply negated its child: a row excluded because its operand was NULL got flipped back in. The two bugs had been cancelling out under `NOT`. Fixing only the comparison trades data loss for wrong results — and no existing test would have caught it.

So the evaluator is made three-valued, per standard SQL:

* `filter_match_t {no, yes, unknown}` replaces `bool` in `column_data_t::check_predicate` and `row_group_t::check_predicate`.
* A NULL operand yields `unknown`, gated **once** at the top of `column_data_t::check_predicate` — so raw buffers, RLE/dictionary, struct/array/list and the update overlay are covered by one check instead of three ad-hoc ones. It routes through `fetch_row`, which applies the update overlay, so a row updated to NULL is seen as NULL.
* `AND`/`OR`/`NOT` propagate UNKNOWN: FALSE dominates AND, TRUE dominates OR, and `NOT UNKNOWN` stays UNKNOWN.
* `IS NULL` / `IS NOT NULL` are always TRUE or FALSE, never UNKNOWN — they are answered from the validity mask before this gate, so they are unaffected.
* Only `yes` selects a row.

## JSONB: the same bug, not a second one

`t ->> 'x'` is resolved at transform time into a plain column reference, and an absent key is stored as a NULL — so an absent-key row reaches the identical scan path. Verified: one guard fixes all three spellings at once.

```sql
CREATE TABLE z.doc ();
INSERT INTO z.doc (id, x) VALUES (1, 5);
INSERT INTO z.doc (id)    VALUES (2);      -- no 'x' key at all
INSERT INTO z.doc (id, x) VALUES (3, 0);

SELECT co ->> 'x' FROM z.doc WHERE id = 2;        -- NULL   (projection was always correct)
SELECT id FROM z.doc WHERE doc ->> 'x' = 0;       -- was: 2 rows, now: 1 (id=3)
DELETE FROM z.doc WHERE doc ->> 'x' = 0;          -- no longer deletes the keyless row
UPDATE z.doc SET id = 99 WHERE doc ->> 'x' > -1;  -- no longer clobbers it
```

This is the `absent_key_coerced_to_zero` bug characterized in #560, but it is a core engine bug that jsonb merely provides another door into — not a jsonb quirk. Note that #560's pins assert the current (broken) behaviour, so they will need flipping once this lands.

## Tests

`integration/cpp/test/test_null_three_valued_logic.cpp` — 9 cases, 110 assertions over both surfaces: comparisons, `NOT`, AND/OR propagation, `DELETE`/`UPDATE`, the update overlay, string columns (empty string must not be conflated with NULL), and the jsonb navs (`->>`, `#>>`, nested `a.b`, plus DML through a nav).

* All 9 cases **fail before** this change (29 failing assertions) and **pass after**. `not_does_not_resurrect_null` exists specifically to fail against the naive validity-guard-only fix.
* Full integration suite on this branch: **410 tests, 0 failures**.

## Known gaps, deliberately not fixed here

Verified identical with and without this change (documented in the test file):

* **Index scan silently drops rows** when the column contains a NULL: with an index on `x`, `WHERE x = 0` returns **0 rows** — losing the row whose value genuinely is `0`. A pure compare on an indexed column is planned as an `index_scan`, bypassing the scan filter entirely. Comparable severity; worth its own fix.
* `UPDATE t SET x = NULL` does not mark the row NULL (`x IS NULL` still does not see it afterwards).
* `WHERE a.b = 0` does not resolve on a computing table (the nested test navigates via `#>> '{a,b}'` instead).

## Follow-up

The validity gate calls `check_validity`, which goes through `fetch_row` and allocates per row on the hot scan path. Correct, but the plain fast path exists to avoid exactly that. Column stats already track `null_count()`, so the clean optimisation is to skip the check on null-free columns — not built on that here, because stale stats would silently reintroduce the bug. Not benchmarked.

### DIFF
--- a/components/table/column_data.cpp
+++ b/components/table/column_data.cpp
@@ -500,13 +500,28 @@ namespace components::table {
         transient.revert_append(static_cast<uint64_t>(start_row));
     }
 
-    bool column_data_t::check_predicate(int64_t row_id, const table_filter_t* filter, core::error_t& error) {
+    filter_match_t column_data_t::check_predicate(int64_t row_id, const table_filter_t* filter, core::error_t& error) {
+        // SQL three-valued logic, enforced once for every column type and every storage
+        // layout. A NULL operand makes any value comparison UNKNOWN — the row is neither
+        // selected nor eligible to be flipped back in by NOT.
+        //
+        // This gate must come first: the compressed and raw-buffer paths below read the
+        // value bytes only, and a NULL row's payload is a meaningless 0 (validity lives in a
+        // separate sibling column), so they would otherwise report `0 == 0` and match.
+        // check_validity() goes through fetch_row, which applies the update overlay, so a row
+        // updated to NULL is seen as NULL here too.
+        //
+        // IS NULL / IS NOT NULL never reach this function — row_group_t::check_predicate
+        // answers them from the validity mask directly — so gating here cannot break them.
+        if (!check_validity(row_id)) {
+            return filter_match_t::unknown;
+        }
         if (updates_ &&
             updates_->has_updates(static_cast<uint64_t>(row_id - start_) / vector::DEFAULT_VECTOR_CAPACITY)) {
             // The vector has some updated rows. Check if THIS specific row is updated.
             if (updates_->row_is_updated(row_id)) {
                 // Row is in the update overlay — check against the updated value only
-                return updates_->check_row(row_id, filter);
+                return updates_->check_row(row_id, filter) ? filter_match_t::yes : filter_match_t::no;
             }
         }
         // STRUCT columns store child data in sub_columns, not in data_ segments — fetch the full value
@@ -522,15 +537,13 @@ namespace components::table {
             fetch_row(fetch_state, row_id, result, 0);
             if (fetch_state.fetch_error.contains_error()) {
                 error = fetch_state.fetch_error;
-                return false;
-            }
-            if (!result.validity().row_is_valid(0)) {
-                return false;
+                return filter_match_t::no;
             }
             if (auto* set = dynamic_cast<const set_membership_filter_t*>(filter)) {
-                return set->contains(result.value(0));
+                return set->contains(result.value(0)) ? filter_match_t::yes : filter_match_t::no;
             }
-            return filter->cast<constant_filter_t>().compare(result.value(0));
+            return filter->cast<constant_filter_t>().compare(result.value(0)) ? filter_match_t::yes
+                                                                              : filter_match_t::no;
         }
         auto segment = data_.get_segment(row_id);
         // For compressed segments, fetch the actual decompressed value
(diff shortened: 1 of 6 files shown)


cleveralbatraoz:
Superseded by #570, which bundles this storage-scan/index three-valued-logic work together with the arithmetic, aggregate, and in-memory-predicate 3VL fixes so the whole NULL story lands as one coherent change (these two commits are included there verbatim). Closing in favour of #570.

PR #589, opened 2026-07-23
scikit-build update

### LINKED ISSUE #586: Scikit-build (classic) 1.0 internals
Hi,

I'm currently working on swapping out scikit-build (classic)'s internals for scikit-build-core, and I have AI running lots of downstream projects and looking for regressions. It found that this project currently uses scikit-build internals. I've been able to patch some of them back in as shims, but otterbrix uses `constants.CMAKE_BUILD_DIR` and/or `set_skbuild_plat_name`, which are no longer supported either. I might be able to shim `CMAKE_BUILD_DIR` if really needed.

See https://github.com/scikit-build/scikit-build/pull/1185. I can try to propose something here if you'd like. The best thing would be to use scikit-build-core without setuptools directly.

OrangeSmokingJacket:
Removed deprecated functionality and switched to scikit-build-core


kotbegemot (approved):

integration/python/otterbrix/pyproject.toml
```diff
@@ -59,4 +47,26 @@ classifiers = [
...
+]
+
+[project.optional-dependencies]
+test = ["pytest"]
+
+# CI copies this file to the repo root and builds there, so source-dir is "." (top-level CMakeLists.txt)
+[tool.scikit-build]
+cmake.version = ">=3.20"
+cmake.source-dir = "."
+cmake.build-type = "Release"
```
henryiii, line 59: cmake.version will be read from your CMakeLists if not given, and the other two are defaults (all of which is fine, just pointing it out!).

I'd recommend `minimum-version = "build-system.requires"`, which will read the `>=0.10` from your requirements and use that to keep scikit-build-core's behavior locked to that version.

PR #594, opened 2026-07-27
Reject non-numeric arithmetic and LIKE operands cleanly

cleveralbatraoz:
Companion to #591/#592: covers the operand-guard gaps those left open, and fixes unary plus, which was silently lowered to negation. No promotion or cast machinery is touched.

### 1. Unary plus negated its operand

The transformer lowered every no-left-operand arithmetic operator to `unary_minus`:

```sql
CREATE TABLE db.t (id INT, v BIGINT, b BOOLEAN, s TEXT);
INSERT INTO db.t (id, v, b, s) VALUES (1, 7, true, 'a'), (2, 3, false, 'b');

SELECT +v FROM db.t;               -- before: -7, -3;  now: 7, 3 (SQLite semantics: +X ≡ X)
SELECT +s FROM db.t;               -- before: "unary minus requires a numeric operand"; now: 'a', 'b'
SELECT id FROM db.t WHERE +v > 5;  -- before: 0 rows (filtered on -v);  now: id 1
SELECT SUM(+v) FROM db.t;          -- before: "single-operand scalar is not supported"; now: 10
```

The SELECT field clause and arithmetic-operand resolution peel `+`-layers and dispatch the stripped node to its own arm; in expression slots that require an expression the identity is encoded as `0 + x`.

### 2. WHERE-clause arithmetic had no operand guards

`WHERE` evaluates per row through the predicate value getters, which #592's evaluator guards do not reach:

```sql
SELECT id FROM db.t WHERE b + 1 > 0;  -- before: silently 0 rows (indeterminate promotion read)
                                      -- now: ERROR arithmetic requires numeric or compatible temporal operands
SELECT id FROM db.t WHERE s + 1 > 0;  -- before: SIGSEGV (swallowed throw, scan-teardown crash)
SELECT id FROM db.t WHERE -s < 0;     -- before: garbage comparison; now: ERROR
```

Guards sit after the NULL early-outs: `WHERE v + 1 > 5` over a NULL row still drops it as UNKNOWN, never errors.

### 3. UPDATE arithmetic silently overwrote data

```sql
UPDATE db.t SET v = b + 1;  -- before: reported success, v became NULL in every row
                            -- now: ERROR, data intact
```

### 4. Mixed boolean/numeric INSERT crashed

The INSERT widening path treated BOOLEAN as numeric and asked the promotion oracle for a `(INTEGER, BOOLEAN)` common type:

```sql
CREATE TABLE db.c ();
INSERT INTO db.c (x) VALUES (1), (true);  -- before: SIGSEGV; now: succeeds, per-value store
```

The widening condition now uses `is_arithmetic_numeric` (from #592); the promotion machinery itself is untouched.

### 5. Tests pinning #591's LIKE rejection

The non-string-pattern rejection is pinned across WHERE / CASE WHEN / JOIN ON, for LIKE / NOT LIKE / ILIKE with integer and float patterns, over typed and computing tables, with string-pattern controls.

Full suite green on this branch; error codes/messages match #592's conventions.

### DIFF
--- a/components/expressions/update_expression.cpp
+++ b/components/expressions/update_expression.cpp
@@ -410,6 +410,17 @@ namespace components::expressions {
             owned_output_.emplace(vector_ops::apply_unary_vector_op(resource, *unary_op, *left_vec, vec_count));
         } else if (auto arith_op = to_arith_op(type_)) {
             auto* right_vec = right_->output_vec();
+            // A NULL literal answers NULL through the kernels' NA path; a pair of concrete
+            // types with no operator entry must not silently overwrite the column with the
+            // kernels' all-NULL unresolvable result.
+            if (left_vec->type().type() != types::logical_type::NA &&
+                right_vec->type().type() != types::logical_type::NA &&
+                types::arithmetic_result_type(left_vec->type().type(), right_vec->type().type(), *arith_op) ==
+                    types::logical_type::NA) {
+                return core::error_t{
+                    core::error_code_t::arithmetics_failure,
+                    std::pmr::string{"arithmetic requires numeric or compatible temporal operands", resource}};
+            }
             if (*arith_op == arithmetic_op::divide || *arith_op == arithmetic_op::mod) {
                 types::logical_value_t zero(resource, right_vec->type());
                 for (uint64_t i = 0; i < vec_count; ++i) {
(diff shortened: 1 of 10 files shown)


OrangeSmokingJacket (approved):

OrangeSmokingJacket:
For some reson logical_value.cpp changes are bloated. (instead of ~40 lines, github shows 2000). Could be line endings or tabs/spaces or something else

PR #621, opened 2026-08-12
fix(sql): name resolution fixes

### LINKED ISSUE #620: Name resolution: a qualifiers are not consulted, `JOIN … USING` is a cross product
A column reference was resolved by trying its first segment as a table, then as a database,
and pushing whatever was left into the key as a path inside the value. Nothing checked that
the qualifier named the relation the column came from, and four- and five-segment forms had
no branch at all. Resolution is split across two layers — the transformer matches a reference
against the FROM elements, the validator matches an unqualified one against the column lists —
and both had the same defect independently.

Separately, `JOIN … USING` and `NATURAL JOIN` parse and run but carry no join condition.

Everything below reproduces on `113437a9`.

---

## Silently wrong data

**1. `JOIN … USING` returns the cross product** — silent

```sql
CREATE TABLE db.a (id INT, av INT);   -- 4 rows
CREATE TABLE db.b (id INT, bv INT);   -- 4 rows, 2 ids in common
SELECT a.av, b.bv FROM db.a a JOIN db.b b USING (id);   -- 16 rows, expected 2
```
The columns land in `usingClause`, `quals` stays NULL, and a join with no `quals` is lowered
to a cross product with an `all_true` predicate. `usingClause` is read nowhere.

**2. `NATURAL JOIN` returns the cross product** — silent

```sql
SELECT * FROM db.a NATURAL JOIN db.b;   -- N×M rows
```
`isNatural` is set by the parser and read by nobody.

**3. `LEFT JOIN … USING` is a cross product wearing a left join** — silent

```sql
SELECT * FROM db.a a LEFT JOIN db.b b USING (id);   -- every left row matches every right row
```
The join type survives, the predicate does not.

**4. A qualifier that matches nothing is ignored, and another relation answers** — silent

```sql
CREATE TABLE db.a (id INT, av INT);
CREATE TABLE db.b (id INT, v INT);
-- `v` belongs to y, not x:
SELECT sub.v FROM (SELECT x.v FROM db.a x JOIN db.b y ON x.id = y.id) sub;   -- returns y.v
-- `bv` belongs to b, not to the derived table s:
SELECT s.bv FROM (SELECT x.id, x.av FROM db.a x) s JOIN db.b b ON s.id = b.id;   -- returns b.bv
```
When a qualifier matches no relation in the schema, the validator falls back to matching by
column name alone.

**5. A subquery on the right of a composite left side is dropped from the plan** — silent

```sql
SELECT … FROM db.a JOIN db.b ON … JOIN (SELECT …) c ON …;   -- c is not in the plan, no error
```
The branch handling a subquery on the right exists only when the left side is a plain table.

**6. `(t.x).f` means different things in SELECT and in WHERE** — silent in WHERE

```sql
SELECT (t.custom_type).f3.f1 FROM db.t WHERE (t.custom_type).f3.f1 = 1;
```
SELECT builds the whole path but drops the database qualifier; WHERE resolves the qualifier
but loses the middle path segment, so the predicate reads `custom_type/f1` — a different field.

---

## Crash

**7. A FROM name of more than four segments crashes the process**

```sql
SELECT id FROM a.b.c.d.e;   -- SIGSEGV
```
The grammar's `default` branch returns a `RangeVar` with every field NULL instead of raising
a syntax error, and the transformer dereferences it.

---

## Legal SQL refused

**8. Four- and five-segment column references do not resolve**

```sql
SELECT id FROM shop.pub.orders;              -- works
SELECT shop.pub.orders.id FROM shop.pub.orders;   -- "path: 'shop/pub/orders/id' was not found"
```
A table can be named in full in FROM but its column cannot.

**9. A subquery on the left of a join is a hard error**

```sql
SELECT … FROM (SELECT …) a JOIN db.t b ON …;
-- "incorrect type for join join->larg node T_RangeSubselect"
```

**10. A table function's alias never reaches the plan**

```sql
SELECT g.g FROM generate_series(1, 3) g;   -- "path: 'g' was not found"
```
The alias is used while resolving the reference and then dropped, so the column keeps the
function's name. Nothing addresses the element by the alias it was given.

---

## Resolved against the wrong thing

**11. A CTE is found by bare name whatever the qualification**

```sql
WITH w AS (SELECT …) SELECT w.c FROM anydb.anysch.w;   -- finds the CTE
```
Qualification does not apply to a CTE name; `FROM d.w` must look for a table `w` in `d`.

**12. A single subquery in FROM is not registered as a FROM element**

```sql
SELECT s.c FROM (SELECT … ) s;
```
Neither the name nor the alias is recorded, so `s.c` resolves by accident downstream rather
than against the FROM clause.

---

## Wrong diagnosis

**13. A column present on both sides of a join is reported as missing**

```sql
SELECT a.av FROM db.a a JOIN db.b b ON a.id = b.id WHERE id > 1;
-- "path: 'id' was not found", expected "is ambiguous"
```
Both sides fail for their own reason and the reason is overwritten with "not found", sending
the reader after a typo in a name that is there.

madoxann:
Standardize name resolution for 4 and 5-part names, fixes name resolution bugs, closes #620

### DIFF
--- a/services/dispatcher/validate_logical_plan.cpp
+++ b/services/dispatcher/validate_logical_plan.cpp
@@ -148,6 +148,13 @@ namespace services::dispatcher {
             return merged;
         }
 
+        [[nodiscard]] core::error_t ambiguous_key(std::pmr::memory_resource* resource,
+                                                  const components::expressions::key_t& key) {
+            return core::error_t(
+                core::error_code_t::ambiguous_name,
+                std::pmr::string{"path: \'" + key.as_string() + "\' is ambiguous. Use aliases or full path", resource});
+        }
+
         [[nodiscard]] core::result_wrapper_t<type_paths> find_types(std::pmr::memory_resource* resource,
                                                                     components::expressions::key_t& key,
                                                                     const named_schema& schema) {
@@ -192,12 +199,22 @@ namespace services::dispatcher {
                     }
                 }
             }
-            // Either unqualified, or the qualifier matched nothing because this schema
-            // carries no alias for it (raw node_data inputs have an empty result_alias):
-            // fall back to the by-name rules, which is what every key used to get. Decided
-            // once, before the loop: testing matches.empty() per iteration would stop at
-            // the first hit and hide the ambiguity that collecting every match detects.
-            const bool match_by_name = matches.empty();
+            // Either unqualified, or the qualifier names nothing this schema knows: a
+            // sub-plan's schema is labelled with the derived table's alias, so a reference
+            // written inside it against the inner relation's own name ('inner_t.k' against
+            // 'sub'.k) matches nothing here, and raw node_data inputs carry no alias at all.
+            // Both are resolved by the by-name rules, which is what every key used to get.
+            // A qualifier the schema does know is the opposite case: the relation is right
+            // here and simply has no such column, and answering with a same-named column of
+            // another relation is a wrong answer, not a fallback. Decided once, before the
+            // loop: testing matches.empty() per iteration would stop at the first hit and
+            // hide the ambiguity that collecting every match detects.
+            const bool schema_knows_qualifier =
+                truncated_key.has_qualifier() &&
+                std::any_of(schema.begin(), schema.end(), [&truncated_key](const type_from_t& entry) {
+                    return core::pmr::operator==(entry.result_alias, truncated_key.qualifier());
+                });
+            const bool match_by_name = matches.empty() && !schema_knows_qualifier;
             for (size_t i = 0; match_by_name && i < schema.size(); i++) {
                 if (truncated_key.storage().size() > 2 &&
                     core::pmr::operator==(schema[i].result_alias, truncated_key.storage().at(1)) &&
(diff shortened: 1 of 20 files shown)


services/dispatcher/validate_logical_plan.cpp
```diff
@@ -148,6 +148,13 @@ namespace services::dispatcher {
             return merged;
         }
 
+        [[nodiscard]] core::error_t ambiguous_key(std::pmr::memory_resource* resource,
+                                                  const components::expressions::key_t& key) {
+            return core::error_t(
+                core::error_code_t::ambiguous_name,
+                std::pmr::string{"path: \'" + key.as_string() + "\' is ambiguous. Use aliases or full path", resource});
+        }
```
OrangeSmokingJacket, line 163: I should update doc for error handing for that case...
error_t takes std::source_location as last argument in debug mode, which is useful, if you want to trace where it was originally created, but here it would always trace to line 153, without reference to original call of ambiguous_key().

I am not sure how useful it is here, but is it something to keep in mind

OrangeSmokingJacket (approved):

PR #627, opened 2026-08-25
Improved integer literals processing

### LINKED ISSUE #561: Silent wrong-result and data-corruption bugs
Statements that succeed but return a wrong answer or corrupt data — no error.

---

### 1. A NULL / absent value is read as `0` in a comparison (WHERE/DELETE/UPDATE)

```sql
CREATE TABLE db.t (id bigint, x bigint);
INSERT INTO db.t (id, x) VALUES (1, 5);
INSERT INTO db.t (id)    VALUES (2);          -- x is NULL
INSERT INTO db.t (id, x) VALUES (3, 0);

SELECT id FROM db.t WHERE x = 0;    -- returns {2,3}; should be {3}
SELECT id FROM db.t WHERE x > -1;   -- returns {1,2,3}; should be {1,3}
DELETE FROM db.t WHERE x = 0;       -- also deletes row 2 (x is NULL)
```

A NULL is coerced to `0` in the comparison, so every predicate that `0` satisfies
wrongly matches the NULL rows — and `DELETE`/`UPDATE` mutate them.

### 2. Chained (3+-table) JOIN: a shared column name resolves to the leftmost table

```sql
CREATE TABLE db.l (k bigint, v bigint);
CREATE TABLE db.m (k bigint, v bigint);
CREATE TABLE db.n (k bigint, v bigint);
INSERT INTO db.l (k, v) VALUES (1, 10);
INSERT INTO db.m (k, v) VALUES (1, 20);
INSERT INTO db.n (k, v) VALUES (1, 30);

SELECT m.v FROM db.l JOIN db.m ON l.k = m.k JOIN db.n ON m.k = n.k;
-- returns 10 (l.v); should be 20 (m.v)
```

With three or more joined tables sharing a column name, a reference to a
middle/right table silently returns the leftmost table's value — the binary
left/right side stamp cannot distinguish three tables.

### 3. Restart + schema-widening INSERT scrambles a named projection

```sql
-- session 1:
CREATE TABLE db.t ();
INSERT INTO db.t (id, p) VALUES (1000, 1111);
-- restart (reopen the same data directory), session 2:
INSERT INTO db.t (id, p, q) VALUES (2000, 2222, 3333);

SELECT id, p, q FROM db.t;   -- row 0 comes back rotated: 1111 NULL 1000
SELECT * FROM db.t;          -- correct: 1000 1111 NULL
```

After reopening a persisted table (disk + WAL), an INSERT that adds a new column
rotates the columns of an explicit named projection. `SELECT *` is correct, and a
same-session widen is fine.

### 4. Boolean column compared to an integer literal is inverted

```sql
CREATE TABLE db.t (id bigint, x boolean);
INSERT INTO db.t (id, x) VALUES (1, true), (2, false);

SELECT id FROM db.t WHERE x = 1;   -- returns 2 (the false row); should be 1
SELECT id FROM db.t WHERE x = 0;   -- returns 1 (the true row);  should be 2
```

`x = true` / `x = false` are correct; only comparison to an integer literal is inverted.

### 5. Integer literal beyond INT32 wraps on INSERT

```sql
CREATE TABLE db.t (v bigint);
INSERT INTO db.t (v) VALUES (3000000000);   -- stored as -1294967296 (32-bit wraparound)
```

The literal is parsed as 32-bit and overflows before it reaches the `bigint` column.

OrangeSmokingJacket:
* Proper parsing for 64-128 bit integers (both signed and unsigned)
* Update catalog to handle unsigned types
* Fixed incorrect behaviour for casts (previously uint128 + int128 were converted to double, which led to incorrect comparisons)


OrangeSmokingJacket:
PR covers parts of the #561 issue that were left unresolved, eventhough it's name is a bit off

PR #636, opened 2026-09-03
Improved expression support

### LINKED ISSUE #604: Four type-coercion arms in validate_logical_plan are unreachable: the gate asks whether a TYPE exists with the COLUMN's name
Four type-coercion arms in `services/dispatcher/validate_logical_plan.cpp` (DATE/duration, DECIMAL, STRUCT, ENUM) never execute.

## Measurement

Under lldb over the whole 621-case integration suite: the gate above them (`ty_exists`) is evaluated **15 691 times**, and all four arms have **hit count 0**.

## Why — the gate asks the wrong question

The gate is `type_visible(it->type_name())`.

- **DATE / DECIMAL.** `operator_resolve_table.cpp:480` stamps a catalog column type's alias with the **column's name**, and for a GENERIC/decimal extension `type_name()` returns that alias. So the gate effectively asks *"is there a registered TYPE whose name equals this COLUMN's name?"* — opening the arm requires registering a type called `amount` or `d`.
- **STRUCT / ENUM.** `type_name()` is the real UDT name here, but the gate additionally needs a resolve-type leaf for that UDT inside the INSERT's own plan, and the transformer emits those only for `CREATE TABLE` / `CREATE TYPE` / casts. Verified with a probe: `CREATE TYPE probe_enum`, a table with enum/struct/numeric/date columns, and even `INSERT ... VALUES ('even'::probe_enum, ROW(1,2), 1.25, '2024-06-01')` leaves `ty_exists == false` with `type_name() == "probe_enum"`.

## Decision needed

Either fix the gate — in which case the arms come alive, and their behaviour needs review before that happens — or delete them as unreachable. Leaving a dead conversion path behind a condition that asks about the wrong thing is the worst of the three, because it reads as coverage that does not exist.

Tests that do exercise them exist at dispatcher level (`services/dispatcher/tests/test_write_set_column_identity.cpp`), which is the only altitude where the arms run at all.

### LINKED ISSUE #611: A keyed GROUP BY over zero rows returns no column descriptors at all (the scalar aggregate shape returns them correctly)
```sql
SELECT k, COUNT(*) AS n FROM empty_table GROUP BY k;
```

returns **no column descriptors at all** — not two columns with zero rows, but an empty schema. A client asking the result for its column names gets nothing back.

The scalar shape is fine: `SELECT COUNT(*), SUM(v) FROM empty_table` answers `{"count", "s"}` with one row, because it goes through a separate emit site (`empty_aggregate_result`). The keyed shape does not: `materialize_groups()` never runs when there are no groups, and nothing else names the output.

## Why it is worth fixing rather than documenting

The plan knows the answer. The aggregate node's output schema carries both column names before execution begins — the emit site simply is not reached. A zero-row result that cannot describe its own columns forces every client to special-case emptiness, and it differs from what the same query shape returns when one row exists.

Related, already fixed: the four scan-side `make_drain_chunk` sites had the same shape (an empty user-visible result built from a bare type list) and now name and identify their columns. A zero-row *scan* answers with proper column descriptors; a zero-row *grouped aggregate* still does not.

## Pinned, deliberately

`integration::cpp::test_group_operator_contracts` — "group operator contracts: output column names" asserts the current empty answer explicitly, so closing this has to change a test that says what it is doing, rather than silently starting to pass.

OrangeSmokingJacket:
Fixes open expression issues that we had


OrangeSmokingJacket:
Before merge and formatting total diff was +2618 -713

Issue #581, opened 2026-07-17
Isolation gaps across engine layers


cleveralbatraoz:
### Join / column binding

Database-qualified column references bind by bare table name, so same-named tables from different databases cross-contaminate joins.

```sql
CREATE DATABASE db1;
CREATE DATABASE db2;
CREATE TABLE db1.t (id BIGINT, a BIGINT);
CREATE TABLE db2.t (id BIGINT, b BIGINT);
INSERT INTO db1.t (id, a) VALUES (1, 100), (2, 200);
INSERT INTO db2.t (id, b) VALUES (1, 111), (3, 333);
SELECT * FROM db1.t JOIN db2.t ON db1.t.id = db2.t.id;
-- expected: 1 row; actual: 4 rows (the ON clause degenerates to an always-true self-comparison)
```

### Catalog / DDL

Catalog uniqueness is not enforced.

```sql
CREATE DATABASE db1;
CREATE TABLE db1.t (a INT);
CREATE TABLE db1.t (b TEXT);
-- expected: error "relation t already exists"; actual: success (second table is created as an unreachable orphan)
```

### Session / connection API

Explicit transactions do not work through the public connection surfaces (C++ `connection_t`, C API, Python): each statement runs in its own internal session.

```sql
CREATE DATABASE db1;
CREATE TABLE db1.t (id BIGINT);
BEGIN;
INSERT INTO db1.t (id) VALUES (1);
ROLLBACK;
SELECT * FROM db1.t;
-- expected: 0 rows; actual: 1 row (the INSERT autocommitted; BEGIN/ROLLBACK landed in unrelated sessions)
```

### MVCC / concurrent constraint enforcement

Constraint checks race between concurrent transactions.

```sql
CREATE DATABASE db1;
CREATE TABLE db1.t (id BIGINT, v BIGINT);
ALTER TABLE db1.t ADD CONSTRAINT uq UNIQUE (id);
-- session A: BEGIN;
-- session A: INSERT INTO db1.t (id, v) VALUES (1, 10);
-- session B: BEGIN;
-- session B: INSERT INTO db1.t (id, v) VALUES (1, 20);
-- session A: COMMIT;
-- session B: COMMIT;      -- no statement reports an error
SELECT COUNT(*) FROM db1.t WHERE id = 1;
-- expected: 1 (one transaction must fail); actual: 2 (duplicate unique key committed, survives restart)
```

### Storage / restart

[...]

PR #656, opened 2026-09-11
Improved transactions handling

### LINKED ISSUE #504: Storing settings
Currently all configuration settings are lost on relaunch. It is possible to save them in the catalog, but I am not sure, that lazy loading is suitable for them.

### LINKED ISSUE #612: pg_attribute.added_at_commit_id is never stamped, and the obvious fix makes ALTER-added columns unresolvable to the next statement
`pg_attribute.added_at_commit_id` is never stamped: the backfill that writes it scans under the transaction-less "see all committed" view, which cannot see the row its own transaction just wrote.

Its sibling `dropped_at_commit_id` had the identical defect and has now been fixed (via a snapshot-scoped scan overload) because VACUUM's column-compaction gate is its first real reader. `added_at` was **deliberately left inert**, and this issue records why, because the obvious fix does not work.

## What happens if you switch it to the transaction snapshot

The patch lands and the value is written — and then a column added by `ALTER TABLE ... ADD COLUMN` becomes unresolvable to the **next statement**:

```
INSERT INTO t (id, name, a, b) VALUES (...)   ->  path: 'a' was not found
```

reproducible via `streaming_ddl_leaf::multi_clause_alter_streams` and `idx_null::add_column_then_create_index`.

## Why the obvious explanation is wrong

`resolve_table` hides a column only when `added_at > snapshot.start_time`. In the failing run the written commit id was **6** and the next statement's start time was **7**, on the one clock both values come from — so the visibility rule should not hide it. Something else about writing that row in place is what breaks resolution, and it needs its own investigation rather than being carried as a rider on the VACUUM work.

## Current state, deliberately visible

The scoping is documented on `read_own_writes` in `update_pg_attribute_commit_id_field_inner`, and `pg_attribute_commit_ids_are_stamped` asserts `row.added_at == 0` — written that way so that closing this gap makes a test fail loudly in the place that explains it, rather than silently starting to pass.

### LINKED ISSUE #616: Failed statement inside an explicit transaction discards earlier statements' delete marks and append slots un-reverted
Found during the review of #615 (which fixes the autocommit half of this family).

When a statement FAILS inside an explicit transaction, the executor's failure path (`revert_failed_txn` in `services/collection/executor.cpp`) ends the txn via plain `txn_abort_msg`, which only calls `txn_manager_.abort(session)` — **no drain**. Earlier successful statements in the same txn have already parked their `dml_appends`/`dml_deletes` on the dispatcher-owned transaction via `txn_accumulate_msg` and cleared `exec_result`, so the failure path only reverts the *failed statement's* ranges. The parked delete marks and physical append slots of the earlier statements are silently discarded un-reverted.

Consequence, e.g.:

```sql
BEGIN;
DELETE FROM t WHERE id = 1;   -- ok: stamps delete marks, parks ranges
DELETE FROM t WHERE id = 2;   -- fails (e.g. FK): whole txn aborts
```

Row 1's delete marks stay stamped with the aborted txn id → the same undeletable-row / UPDATE-duplicates-row corruption as #552, plus leaked physical append slots for any earlier INSERT/UPDATE in the txn.

The explicit ROLLBACK path handles this correctly (`operator_abort_transaction` drains via `txn_abort_drain_msg` and reverts the union). The failed-statement path should go through the same drain + revert instead of the bare `txn_abort_msg`.

OrangeSmokingJacket:
This PR fixes most issues related to a single session with multiple transactions:
* Removed old untransactional methods
* autocommit option added
* DDL changes can be viewed within the same transaction
* errors within transaction cause a ROLLBACK (including parser errors)
* new transactions for the same session are queued in transaction manager
* And other small fixes and improvements

What is still unresolved:
* register/unregister cast/function -> registries do not have MVCC built in, so they can not track different versions on the fly
* same issue with indexes -> no MVCC means that there is no way to track progress between different transactions
* Concurrency between different sessions


OrangeSmokingJacket:
Some of the sub issues of #581 are fixed:
* single session errors
* table/column name uniqueness

PR #658, opened 2026-09-18
fix(sql): name resolution followups

### LINKED ISSUE #574: Name resolution follow-up gaps
Follow-ups blocked on #573 (namespace-aware table resolution) or on a default-database design.

### 1. Duplicate CREATE TABLE silently succeeds

No `resolve_table` sibling is emitted for the create target, so `check_collection_exists` consults an empty resolve index and always reports "not exists"; a second pg_class row is appended. Blocked on #573: with relname-only resolution the check would false-positive across databases.

```sql
CREATE DATABASE db1;
CREATE TABLE db1.t1 (id BIGINT);
CREATE TABLE db1.t1 (id BIGINT);  -- OK today; should be "already exists"
```

### 2. Unqualified names have no default database

No session current-database / search-path exists. Unqualified `CREATE TABLE` persists `pg_class.relnamespace = 0`; `gather_plan_resolve_index` drops metadata whose namespace is INVALID, so unqualified DML fails validation. Needs a default-namespace design (PostgreSQL: current database + `search_path`), including the meaning of qualified access to tables created unqualified and migration of existing `relnamespace = 0` rows.

```sql
CREATE TABLE t1 (id BIGINT);      -- OK, persists relnamespace = 0
INSERT INTO t1 (id) VALUES (7);   -- error: collection does not exist
SELECT * FROM t1;                 -- error: collection does not exist
```

### 3. Same-named UDFs alias across databases

`resolve_function_by_name` filters pg_proc by `proname` only (`services/disk/manager_disk_resolve.cpp:81-111`); `pronamespace` is written at registration but ignored at lookup, so two same-named functions in different namespaces resolve first-match-wins — the same pattern #573 fixed for tables. Unqualified calls need the default-namespace design from item 2 to know which namespace to filter by.

madoxann:
Follow-ups name resolution issues in DML & UDF/Index creation. Closes #574


components/physical_plan/operators/operator_register_udf.cpp
```diff
@@ -16,8 +17,19 @@
...
+        bool function_exists(const std::string& name) {
+            const auto* registry = components::compute::function_registry_t::get_default();
+            if (registry == nullptr) {
+                return true;
+            }
+            const auto functions = registry->get_functions();
+            return std::any_of(functions.begin(), functions.end(), [&name](const auto& function) {
+                return function.first == name;
+            });
+        }
```
OrangeSmokingJacket: It is dangerous to rely on default registry.
Any user defined function will be excludeed there, and whole point of that registry is to support them in the first place

OrangeSmokingJacket (approved):

OrangeSmokingJacket (approved):

